Issue #57, opened 2026-04-22
poc_example fault injection fails on GitHub-hosted ubuntu; passes in Lima


avatar29A:
Discovered on the first CI run of PR #55 (testops harness).

## Symptom

On \`ubuntu-latest\` (GitHub-hosted runner, kernel 6.x), running the testops harness's \`poc_example\` case:

\`\`\`
--- FAIL: test_db_slow (209ms, seed=1) ---
  reason: assert_eq failed: 0 != 200

--- FAIL: test_happy_path (10208ms, seed=1) ---
  reason: tcp send failed: read: read tcp 127.0.0.1:34466->127.0.0.1:33125: i/o timeout
\`\`\`

The same seed + spec + binaries passes cleanly in Lima (Ubuntu 24.04, kernel 6.8.0, aarch64). Faultbox's setup logs look identical up to \`service_started\` — PID/MNT/USER namespaces install, seccomp filter loads, target execs. Failures appear only when the real client inside \`mock-api\` tries to reach \`mock-db\` via loopback TCP.

## Impact

Blocks most Critical-tier rows in [docs/feature-manifest.md](../blob/main/docs/feature-manifest.md) from reaching green in CI — fault injection, real-syscall temporal assertions, and the poc_example end-to-end path all depend on the seccomp sandbox working on shared runners.

## Suspected causes, in order of likelihood

1. **AppArmor / container profile:** GitHub-hosted runner restricts the network or user-namespace sandbox that faultbox sets up, making in-sandbox services unable to accept localhost TCP from the outer harness.
2. **sysctl difference:** \`/proc/sys/kernel/unprivileged_userns_clone\` or related differs, causing the USER namespace setup to produce a subtly different mapping that breaks loopback reachability.
3. **Timing:** CI runners are I/O-noisier than Lima; 500ms-delay fault plus healthcheck window may not leave enough budget for the real HTTP round-trip on \`test_db_slow\`.

## Repro

Push any branch with \`poc_example\` un-skipped in \`testops/harness.go\`:

\`\`\`
# in testops/harness.go, remove the Skip: line from the poc_example Case
git push
# Default CI workflow reproduces within 15s.
\`\`\`

## Acceptance

\`go test ./testops/... -run poc_example\` passes on \`ubuntu-latest\` GitHub-hosted runner. Likely fixes: running faultbox under sudo on CI; loosening the namespace config for shared-runner environments; documenting the required runner configuration.

avatar29A:
[...]

PR #63, opened 2026-04-22
experiment: relax AppArmor userns restriction for #57

avatar29A:
Experiment to confirm or refute the hypothesis in #57.

## What this does

- Un-skips \`poc_example\` in the testops corpus.
- Adds one CI step: \`sudo sysctl -w kernel.apparmor_restrict_unprivileged_userns=0\` before the testops goldens step.

## Hypothesis under test

Ubuntu 24.04's default \`kernel.apparmor_restrict_unprivileged_userns=1\` applies an AppArmor profile to processes inside unprivileged user namespaces, restricting some network operations. GitHub-hosted \`ubuntu-latest\` runners are Ubuntu 24.04 and inherit this default. Faultbox creates per-service unprivileged USER + PID + MNT namespaces, so this policy applies to the services under test — breaking loopback TCP between them.

## Outcomes

- **Green CI** → hypothesis confirmed. The sysctl becomes a documented CI prerequisite. Keep #57 open for the proper fix: a per-service \`namespaces=\` kwarg that lets specs opt out of USER namespace, making this sysctl unnecessary.
- **Red CI** → hypothesis refuted. Re-skip \`poc_example\`, reopen #57 investigation.

## Not a long-term fix

The sysctl is a system-wide escape hatch. On shared CI, it's acceptable (the runner is ephemeral). On a customer's laptop or security-sensitive CI it isn't — hence the follow-up \`namespaces=\` kwarg.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,15 @@ jobs:
       - name: Build poc binaries for testops
         run: make testops-prep
 
+      # Experiment for issue #57 — Ubuntu 24.04's AppArmor default
+      # restricts processes inside unprivileged user namespaces, which
+      # breaks faultbox's service sandbox on GitHub-hosted runners.
+      # If this makes poc_example pass, the hypothesis is confirmed
+      # and this becomes a documented CI prerequisite (pending a
+      # proper per-service `namespaces=` kwarg).
+      - name: Relax AppArmor userns restriction (experiment for #57)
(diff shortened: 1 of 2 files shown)


avatar29A:
Experiment complete — hypothesis refuted. Sysctl `kernel.apparmor_restrict_unprivileged_userns=0` did not fix poc_example on GitHub runners. See https://github.com/faultbox/Faultbox/issues/57#issuecomment-4295645810 for what we learned and next hypotheses (amd64 arch tables, proxy timing, shared-fd + namespace interaction).

Issue #61, opened 2026-04-22
poc/demo: test isolation leaks proxy/port state between tests


avatar29A:
Discovered while verifying the fix for #56.

## Symptom

When running the full `poc/demo/faultbox.star` spec in one invocation:

\`\`\`
./bin/faultbox test poc/demo/faultbox.star --seed 1
\`\`\`

Only the first test to run (alphabetically: \`test_disk_full\`) passes. All subsequent tests fail with:

- \`assert_eq failed: 0 != 200\` — HTTP response status is 0 (no response body)
- \`fault rules were installed but no injections fired\` warning
- inventory service trace shows only 5 events: \`[openat:4 write:1]\` — inventory starts but never handles a real request

Each test runs in ~200ms (vs 222ms for test_happy_path alone when it passes cleanly), suggesting services come up but requests never round-trip through the faultbox TCP proxy.

## Repro

\`\`\`
# From a fresh environment (no lingering /tmp/inventory.wal or zombie processes):
rm -f /tmp/inventory.wal
sudo pkill -9 -f 'order-svc|inventory-svc' 2>/dev/null
./bin/faultbox test poc/demo/faultbox.star --seed 1
\`\`\`

All 5 tests after \`test_disk_full\` fail with status=0 / proxy-unreachable patterns.

**Crucially: running each test individually passes.**

\`\`\`
for t in test_happy_path test_inventory_slow test_inventory_unreachable \\
         test_wal_fsync_failure test_flaky_network; do
  rm -f /tmp/inventory.wal
  ./bin/faultbox test poc/demo/faultbox.star --test \$t --seed 1
done
\`\`\`

All pass in isolation.

## Suspected cause

faultbox inserts a TCP proxy between orders and inventory (\`proxy_started passthrough mode\` in the trace). The proxy likely doesn't release its listener port cleanly between tests, so the next test's proxy can't bind or route correctly. Orders's connects then fail silently, producing status=0.

Alternative hypothesis: shared state inside the runtime (filter rules, service sessions) accumulates across tests and corrupts later runs.

## Impact

Blocks un-skipping \`poc_demo\` in the testops regression corpus (see PR #55). Also misleads anyone running the demo spec end-to-end — they see 5/6 failures and assume the product is broken.

## Acceptance

\`./bin/faultbox test poc/demo/faultbox.star --seed 1\` passes all 6 tests back-to-back in one invocation, with no manual cleanup between tests. Verify determinism across 5 consecutive runs.

## Related

[...]

PR #65, opened 2026-04-22
diagnostic: amd64 trace capture for #57 investigation

avatar29A:
Diagnostic run to gather concrete evidence for #57. Adds an always-run CI step that prints kernel/AppArmor state and runs poc_example test_happy_path with --debug + console logging. Will close this PR after inspecting the CI logs.

Includes the #61 fix (cherry-picked from #64) so test isolation doesn't muddy the signal. Not intended to merge.


avatar29A:
Proved #57 = #61. PR #64 fixes both.

Issue #94, opened 2026-05-01
RFC-037: Determinism for Remote Services


avatar29A:
> **Status: Draft (problem statement + options — design not yet committed.)**
> Companion to RFC-036 (#91). Split out from the original RFC-036 draft because the design space is large enough to warrant its own discussion thread. We intentionally did not gate RFC-036 on resolving this — the `remote=` primitive is useful even with best-effort reproducibility, and shipping it lets us learn from real customers what their actual determinism needs are.

In-tree RFC document: [`docs/rfcs/0037-remote-determinism.md`](https://github.com/faultbox/Faultbox/blob/main/docs/rfcs/0037-remote-determinism.md).

## Summary

[RFC-036 (#91)](https://github.com/faultbox/Faultbox/issues/91) introduces `service(remote=...)` — a service whose backing process lives in a customer's k8s dev cluster, reached through Faultbox's existing protocol proxy. That primitive is useful on its own, but a remote pod is **shared, mutable, and outside our control**, so runs against it violate the [`.fb` bundle reproducibility contract from RFC-025](https://github.com/faultbox/Faultbox/issues/59) ("one bundle, fully reproducible run anywhere").

This RFC **frames the problem and lays out the design options** for restoring determinism. It does not yet commit to an implementation. The explicit goal is to gather input on which option (or combination) fits Faultbox's positioning best, with at least one customer running on RFC-036 v1 to ground the discussion.

### What problem does this solve?

Faultbox's reproducibility contract is a core differentiator ("not just chaos engineering — every failure is reproducible"). RFC-025 committed us to "every `faultbox test` run drops a single `.fb` file that, on any other machine, can be replayed to produce a byte-equivalent trace." Every primitive shipped since (replay, lock, report, drift warning) builds on that contract.

RFC-036 lets a `service()` declaration point at an externally-running pod. The pod can:

- Return different responses to the same request between runs (state mutates, feature flags flip, time-based responses).
- Be redeployed mid-test-suite.
- Be unreachable from a developer's laptop the next day.
- Return bytes that aren't captured anywhere today.

A bundle produced by a run against `remote=` therefore **violates the RFC-025 contract**. `faultbox replay <bundle>` on a different machine will either re-dial the remote (different bytes back, or connection refused) or fail entirely.

### Why is this important now?

- We're shipping RFC-036 in v0.13.0; the gap is real and visible the moment a customer uses `remote=`.
- Customer demand for `remote=` is real (DevPlatform / inDrive use case).
[...]

PR #99, opened 2026-05-02
RFC-036: Remote services — service(remote=...) for k8s dev environments

### LINKED ISSUE #91: RFC-036: Remote Services — Faulting Dependencies in Customer k8s Dev Environments
> **Update 2026-05-01:** the original draft of this RFC included a record-and-replay design to preserve `.fb` bundle reproducibility against remote services. That's been **split out into RFC-037 (Determinism for Remote Services)** because the design space is large enough to warrant its own discussion thread. This RFC ships the primitive; RFC-037 decides how runs against it are made deterministic.

## Summary

Add a fourth source for `service()` alongside `binary=` / `image=` / `build=`: a **`remote=`** kwarg that points at an externally-running endpoint (a k8s `Service` hostname, a port-forwarded address, an IP). Faultbox does not launch the service; it stands up its existing protocol proxy in front of each declared interface, dials the remote upstream, and lets the SUT exercise it through the proxy unchanged.

Everything that already runs on the proxy data path keeps working — `response()`, `error()`, `slow()`, gRPC-method targeting, SQL matchers, event emission. Faults that require process control — syscall-level `deny()` / `delay()` / `hold()`, `seed=` / `reset=` / `reuse=` — are **rejected at spec-load** with a clear error pointing the user to the protocol layer or to `mock_service()`.

**Out of scope of this RFC:** how runs against a remote service are made deterministic / reproducible. A remote pod is shared, drifts between runs, and isn't captured by the `.fb` bundle (RFC-025) — that's a real regression to Faultbox's reproducibility contract, but the design space (record-and-replay, snapshotting, contract pinning, drift detection) is big enough to warrant its own RFC. **See RFC-037.** This RFC ships the primitive; runs against `remote=` services are explicitly best-effort on reproducibility until RFC-037 lands.

Target release: **v0.13.0**.

In-tree RFC document: [`docs/rfcs/0036-remote-services.md`](https://github.com/faultbox/Faultbox/blob/main/docs/rfcs/0036-remote-services.md).

### What problem does this solve?

inDrive-shape teams keep running into the same wall. Their SUT (`truck-api`) has dependencies — `geo-config`, `pricing`, `auth-server`, `feature-flags`, `dispatch`, `payments`, ... — that are real services maintained by other teams, deployed continuously to a shared k8s dev namespace, and **not distributed as Docker images**. Reasons vary: builds are 8 GB and live on internal artifact stores; private base images; licensed binaries; or simply that nobody on the platform side considered "a developer pulls this and runs it" a supported workflow.

Today the developer's only options are:

1. **`mock_service()` for each dependency.** Faithful for trivial stubs, drifts the moment the real service ships a new contract. Authoring cost grows linearly with the dep graph; many dependencies have 50+ endpoints.
2. **Skip the test.** The thing the customer most wanted to test — *"what happens to truck-api when the real `geo-config` returns 503"* — is the thing they can't express.
3. **Lobby the platform team.** Out of scope for an engineering tool.

### Why is this important now?

This is the gap that the [2026-04-22 customer-feedback analysis](https://github.com/faultbox/Faultbox/blob/main/docs/design/2026-04-22-customer-feedback-analysis.md#L271) explicitly deferred ("DevPlatform integration → 1.x"). Three pieces in tree make it almost-free now:

| Piece | What it gives us |
|---|---|
| `mock_service()` (RFC-017) | Proves the type system already accepts a service with interfaces but no image/binary. Fault rules, events, assertions all work without process attachment. |
| Pre-started proxy datapath (RFC-024) | Every proxy-capable interface already gets a local listener that forwards to "the upstream." Today the upstream is a container we launched. Repointing it is one `Dial`. |
| `seccomp=False` opt-out ([builtins.go:280](https://github.com/faultbox/Faultbox/blob/main/internal/star/builtins.go#L280)) | Establishes the precedent that some services run protocol faults only. Same validation pattern. |

What's missing is a single kwarg on `service()` and the validation that protects users from configurations that can't possibly work.

### What happens if we don't do this?

- Customers continue authoring full mock surfaces for evolving services and shipping drift bugs into their fault matrices.
- The "DevPlatform" lane stays deferred; large k8s shops can't onboard.

### Current State

`service()` requires exactly one of `binary=` / `image=` / `build=`. Validation in [`internal/star/builtins.go:292`](https://github.com/faultbox/Faultbox/blob/main/internal/star/builtins.go#L292) errors otherwise. Every interface gets a pre-started proxy whose upstream is the launched container. Syscall faults install a seccomp BPF filter on that container's process. `mock_service()` (RFC-017) sidesteps the launch and the filter but has no upstream — the mock generates responses itself.

There is no way today to point a `service()` at an externally-running pod.

#### DSL — one new kwarg

```python
geo = service("geo-config",
    interface("public",   "http", 8080),
    interface("internal", "grpc", 9090),
    remote      = "geo-config.staging.svc.cluster.local",
    healthcheck = http("geo-config.staging.svc.cluster.local:8080/healthz"),
)

api = service("truck-api",
    interface("main", "http", 8000),
    image       = "truck-api:dev",
    depends_on  = [geo],
    env         = {"GEO_CONFIG_URL": "http://%s/" % geo.public.addr},
)
```

Behaviour at session start:

1. `service("geo-config", remote=...)` registers a `ServiceDef` with no launch source — same data shape as `mock_service()`.
2. The proxy manager pre-starts a listener on `127.0.0.1:NNN` per interface (RFC-024 path). Upstream addr = `<remote>:<iface.port>` for each interface.
3. `buildEnv` / `buildContainerEnv` substitute the proxy address into every consumer's env exactly as today — no consumer-side change.
4. The remote healthcheck (run on the host, not against the proxy) gates session start. Connection failure surfaces a setup hint.
5. No seccomp filter installs. No container/binary lifecycle. No `seed`, `reset`, `reuse`.

#### Kwarg surface

| Kwarg | On remote services | Notes |
|---|---|---|
| `remote=` | **Required exactly one of** binary/image/build/remote | Plain `host` (interface ports added) or `remotes({...})` per-interface override |
| `interface(...)` | Required | Same as today |
| `healthcheck=` | **Required** | We can't `wait_for_listening_socket()` on a remote pod we don't own; the spec must declare what "ready" means |
| `depends_on=` | Allowed | Topological sort sees `remote` services as roots |
| `env=` | Allowed | Values like `geo.public.addr` resolve to the proxy address, same as today |
| `seccomp=` | Rejected | No process to filter |
| `seed=` / `reset=` / `reuse=` | Rejected | We don't own the lifecycle |
| `volumes=` / `ports=` / `args=` / `binary=` / `image=` / `build=` | Rejected | All meaningless without a launch path |
| `observe=` | Restricted: `stdout()`/`tail()` rejected; `topic()`/`poll()` allowed |

#### Spec-load validation

When a `fault()` rule targets a remote-service interface:

- **Protocol-layer rules** (`response`, `error`, `slow`, gRPC method, SQL matcher, body match) — accepted, dispatched to the proxy identically to a local container.
- **Syscall-layer rules** (anything needing a seccomp filter) — **rejected at spec load** with a clear message:

  ```
  fault.geo-config.public.write: service "geo-config" is remote
    (remote="geo-config.staging.svc.cluster.local"); syscall-level faults
    are not available on remote services. Use a protocol fault
    (response=, error=, slow=) at the interface layer, or replace
    `remote=` with `mock_service()` if you need full control.
  ```

#### Proxy lifecycle change

`internal/star/runtime.go::startServices` learns one branch:

```go
if svc.IsRemote() {
    if err := healthcheckRemote(svc); err != nil { return wrapHint(err) }
    proxyMgr.PreStartForRemote(svcName, svc)
    return nil   // skip launch, skip seccomp, skip seed/reset
}
```

`proxy.Manager.PreStartForRemote` is the same code path as the existing `EnsureProxy` with the upstream addr coming from `iface.RemoteHost:iface.Port`. **No protocol plugin changes.**

A `service_started` event is emitted with `kind="remote"` and the upstream addr in the payload.

#### Connectivity — Faultbox stays cluster-agnostic

The `remote=` value is a plain hostname. Whether that hostname resolves is the user's responsibility. Supported workflows in order of preference:

1. **Telepresence connect** (recommended). User runs `telepresence connect` on their workstation; cluster Service DNS resolves on the host; Faultbox's proxy dials normally. No changes to Faultbox.
2. **In-cluster execution.** If the customer runs `faultbox test` in a pod inside the cluster (CI-on-cluster, dev container), Service DNS works natively.
3. **kubectl port-forward / kubefwd.**
4. **VPN.**

Healthcheck failure on session start emits a hint pointing at these workflows. We considered shipping `telepresence connect` orchestration inside Faultbox; rejected — adds a runtime dependency, conflicts with users who already have a connection up, and ties our binary to Telepresence's release cadence.

#### Optional stdlib helper

Ship `@faultbox/discovery/k8s.star` (RFC-019 distribution pattern) as pure string sugar:

```python
load("@faultbox/discovery/k8s.star", "k8s")

geo = service("geo-config",
    interface("public", "http", 8080),
    remote = k8s.service("geo-config", namespace = "staging"),
)
```

`k8s.service(name, namespace)` returns `"<name>.<namespace>.svc.cluster.local"`. **No runtime k8s client, no kubeconfig parsing in Faultbox itself.**

### Reproducibility — deferred to RFC-037

Remote services break the `.fb` bundle reproducibility contract from RFC-025: a remote pod is shared, drifts between runs, and isn't captured in the bundle today. That's a real regression and we need to fix it — but the design space (record-and-replay, contract pinning, schema snapshotting, drift detection, sensitive-data handling) is big enough that it deserves its own RFC and shouldn't gate landing the primitive. **See RFC-037.**

For v1 of this RFC, runs against `remote=` services are explicitly best-effort on reproducibility:

- The bundle records *that* a remote was used (in `env.json` — `remotes: [{svc, iface, host, resolved_at}]`) so replay attempts can detect "this run used a remote and we can't reconstruct it offline".
- `faultbox replay <bundle>` against a bundle whose `env.json` declares remotes prints a clear warning: *"this run used N remote services; replay will attempt to re-dial them; for offline replay see RFC-037."*
- Customers who care about reproducibility can already swap `remote=` for `mock_service()` and check that into git.

The proxy hook surface for "every interaction with a remote upstream" is the same surface RFC-024 exposes for protocol-rule matching — no further plumbing is needed in this RFC to keep RFC-037's options open.

### Alternatives Considered

1. **Make `remote=` an interface-level kwarg.** Rejected: forces every interface to repeat the host string; conceptually the *service* is remote, the interfaces are how you reach it. Per-interface override (`remotes({...})`) covers the rare split case.
2. **New top-level `remote_service()` builtin.** Rejected: redundant with `mock_service()` already existing as a sibling of `service()`. Keeping `service(remote=...)` mirrors the launch-source kwargs and reuses every downstream concept.
3. **mirrord-style network-namespace injection.** Rejected: requires kernel work on the customer's hosts, conflicts with our seccomp story, and doesn't help the central problem (the customer still can't fault `geo-config`'s responses).
4. **Faultbox-native cluster runner.** Rejected for v1; this is the "hosted runner" deferred to 1.x.

## Impact

- **Breaking changes:** None for existing specs. New kwarg, new validation paths only.
- **Migration:** N/A. Customers adopt `remote=` per-service when they want it.
- **Performance:** Same as today's pre-started proxy (RFC-024) — one extra TCP connection per request to the remote upstream. No seccomp install cost for remote services.
- **Security:** `remote=` proxies still bind on `127.0.0.1` (binary mode) or the Docker network (container mode) — no new exposure. (Sensitive-data handling for recording is a RFC-037 concern.)
- **Reproducibility:** Net negative *until RFC-037 lands*. Documented limitation; visible in `env.json` + `faultbox replay` warning.
- **Positioning:** Unblocks the deferred "DevPlatform integration" lane.

## Open Questions

1. **Healthcheck mandatoriness.** Is requiring `healthcheck=` on remote services a usability tax, or the right pressure on users to declare ready-criteria? Lean toward required.
2. **`remotes({...})` syntax vs flat host:port string.** For services where every interface lives on the same host, `remote = "geo.staging:0"` is shorter than `remotes({...})`. Worth a sugar pass before locking the typed form.
3. **Connection-failure during a run** (vs at startup): the remote pod stays healthchecked at start but goes away mid-test. Do we (a) treat as an unhandled error, (b) emit a synthetic protocol error and let oracles handle it, or (c) abort the run? Lean (b).
4. **Determinism story** — entirely RFC-037's problem; flagging here so reviewers know to tag that RFC for the substantive determinism discussion.

## Implementation Plan

**Phase 1 — Plumbing (v0.13.0)**
- `internal/star/builtins.go`: accept `remote=` kwarg; extend "exactly one source" validation; reject incompatible kwargs on remote services.
- `internal/star/types.go`: `ServiceDef.Remote string` + `RemotePerInterface map[string]string`.
- `internal/star/runtime.go::startServices`: branch for `IsRemote()` — skip launch, run remote healthcheck, call `proxyMgr.PreStartForRemote`.
- `internal/proxy/manager.go`: `PreStartForRemote(svc, iface)` reusing `EnsureProxy` with a remote upstream addr.
- `internal/proxy/manager.go` + plugin loop: validate the syscall-fault rejection at fault registration time.
- `recipes/discovery/k8s.star`: `k8s.service(name, namespace)` helper.

**Phase 2 — Bundle annotation + UX (v0.13.x)**
- `internal/bundle/env.go`: record `remotes: [...]` array in `env.json`.
- `cmd/faultbox/replay.go`: warn when a bundle declares remotes; reference RFC-037.
- `internal/report`: top-of-report annotation when remote services were used.

**Phase 3 — Documentation**
- `docs/spec-language.md`: new "Remote services" section.
- `docs/tutorial/`: chapter walking truck-api through the cluster scenario.
- `docs/guides/connectivity.md`: Telepresence / port-forward / in-cluster recipes.

## Regression Tests (release blocker — must land with v1)

Goal: pin every behaviour the RFC promises so future refactors (and RFC-037 changes) can move fast without re-litigating them. Anything in this list missing on the merge candidate blocks the release.

### Spec-load validation (Starlark layer)

- `TestRemote_RejectsBinaryAndRemote` / `_ImageAndRemote` / `_BuildAndRemote` — `service()` errors at load time naming both kwargs.
- `TestRemote_RequiresHealthcheck` — `service(remote=...)` without `healthcheck=` errors with the documented message.
- `TestRemote_RejectsSeedResetReuse` — each of `seed=` / `reset=` / `reuse=` on a remote service errors and names the offending kwarg.
- `TestRemote_RejectsVolumesPortsArgsSeccomp` — table-driven test asserting every meaningless-on-remote kwarg fails at load with a stable message.
- `TestRemote_AcceptsTopicAndPollObserve` — `observe=[topic(...)]` / `[poll(...)]` accepted; `[stdout()]` / `[tail(...)]` rejected.
- `TestRemote_PerInterfaceRemotesValidates` — `remotes({...})` keys must match declared interface names.
- `TestRemote_FaultRule_RejectsSyscallTarget` — registering a syscall fault against a remote interface errors with the documented multi-line hint pointing to `response()` / `error()` / `slow()` / `mock_service()`.
- `TestRemote_FaultRule_AcceptsProtocolTarget` — `response()`, `error()`, `slow()`, gRPC-method, SQL-matcher, body-matcher all install cleanly.

### Runtime / proxy lifecycle

- `TestRemote_NoSeccompFilterInstalled` — verify no BPF filter is built for any remote service.
- `TestRemote_NoLaunchAttempted` — Docker mock asserts `ContainerCreate` / binary fork are never called for a remote service.
- `TestRemote_PreStartProxyDialsRemote` — `proxy.Manager.PreStartForRemote` binds a local listener; first connection through it dials `remote:iface.port` (verified against an `httptest.Server` stand-in).
- `TestRemote_HealthcheckGatesStartup` — unreachable `remote=` aborts session start with the documented hint string.
- `TestRemote_ServiceStartedEvent_KindRemote` — emitted event's `kind` field is `"remote"` and payload contains the upstream addr.
- `TestRemote_BuildEnvSubstitutesProxyAddr` — consumer env values containing `<remote>:<port>` get rewritten to the proxy addr (binary mode and container mode, both via `consumerMode` parameter from RFC-024 v0.9.6).
- `TestRemote_BundleEnvJSONRecordsRemotes` — `env.json` after a run contains a `remotes: [{svc, iface, host, resolved_at}]` array.

### End-to-end (golden / integration)

- `TestE2E_RemoteHTTP_PassThrough` — fake HTTP server stands in for the remote pod; SUT reaches it via the proxy; bundle records the interaction shape.
- `TestE2E_RemoteHTTP_FaultRewrite` — same setup with `error(path="/v1/regions/**", status=503)`; SUT sees 503; remote fake server logs only the request.
- `TestE2E_RemoteGRPC_FaultRewrite` — same shape for gRPC method-level targeting.
- `TestE2E_RemoteUnreachable_StartupHint` — `remote=` points at an unreachable port; session-start error contains the Telepresence/port-forward/in-cluster hint block verbatim.
- `TestE2E_RemoteMidRunDeath` — fake server dies mid-test; verifies the Open-Question-3 lean-(b) behaviour (synthetic protocol error reaches oracle).
- `TestE2E_RemoteVsLocal_ParityFaultMatrix` — same fault matrix authored against a local container vs a fake-remote endpoint produces bit-equivalent fault outcomes (modulo timing); locks in the "swap one keyword" success criterion.
- `TestE2E_ReplayWarning_RemoteBundle` — `faultbox replay` against a bundle whose `env.json` declares remotes prints the documented warning (regex-matched against stderr).

### Stdlib helper

- `TestK8sDiscovery_ServiceHelper` — `k8s.service("name", namespace="ns")` returns `"name.ns.svc.cluster.local"`; rejects empty name/namespace with a clear error.

### Documentation regression

- `TestDocs_SpecLanguageRemoteSection` — `docs/spec-language.md` contains a "Remote services" subsection (string-grep gate so we don't ship the feature without docs).
- `TestDocs_TutorialRemoteChapter` — equivalent gate on `docs/tutorial/`.
- `docs/feature-manifest.md` — `remote=` row appears with greenlight tier and links to the test files above (release-gate per the `/release` skill — RFC-036 entries must exist before tagging).

### Pre-release sanity

- `make test` clean on linux/arm64 + linux/amd64.
- `make demo-container` covers a `remote=` flow against a side-launched fake (no Lima-only paths; fake-remote runs as a sibling container so the regression survives without Telepresence/cluster access).
- `go vet ./...` clean — including new `internal/proxy` and `internal/star` files.
- `.fb` bundle written by a remote-using run round-trips through `faultbox inspect` cleanly.

### What this catches

The list locks down every behaviour the RFC promises: the DSL surface (which kwargs error / which work), the runtime contract (no launch, no seccomp, healthcheck-gated start), the data path (proxy actually forwards, env actually rewrites), the artifact contract (`env.json` shape, replay warning), and the documentation gate. Deliberately does *not* cover RFC-037 territory (recording correctness, replay determinism) — those tests land with that RFC.

## Dependencies

- **Depends on:** RFC-017 (mock services), RFC-024 (proxy datapath).
- **Companion:** RFC-037 (Determinism for Remote Services) — separate RFC, separate issue.
- **Unblocks:** the deferred "DevPlatform integration" lane (FB §6.3 #9).

## Non-goals

- **Faultbox does not run a k8s control plane.** No CRDs, no operator, no in-cluster install.
- **Faultbox does not orchestrate Telepresence.** Documented as the recommended setup, not invoked by us.
- **Faultbox does not intercept inbound traffic to remote pods.** That's `telepresence intercept` territory.
- **No mirrord-style pod injection.**
- **No deterministic replay of remote interactions.** Out of scope; see RFC-037.

avatar29A:
## Summary

Implements [RFC-036](https://github.com/faultbox/Faultbox/issues/91) — `service(remote=...)` for pointing Faultbox at externally-running endpoints (k8s pods, port-forwards, IPs) instead of launching a process. Faultbox stands up its existing protocol proxy in front of each interface and dials the remote upstream; protocol-level faults work unchanged via RFC-024's data path. Process-level kwargs (`seed`, `reset`, `reuse`, `volumes`, `ports`, `args`, `seccomp`, `observe`, `ops`) and syscall-level faults (`write=deny()`, `connect=delay()`, etc.) are rejected at spec load with explicit messages pointing at protocol faults or `mock_service()`.

The companion [RFC-037](https://github.com/faultbox/Faultbox/issues/94) drafts the determinism story (record-and-replay vs contract pinning vs explicit non-determinism marker) — intentionally out of scope here. This PR records remotes used in `env.json` and `faultbox replay` warns; offline-deterministic replay lands in a follow-up.

## Test plan

- [x] `go test ./...` clean across the repo
- [x] `go vet ./...` clean
- [x] `go build ./...` clean
- [x] 32 spec-load validation tests (every accepted/rejected kwarg, fault rule routing, per-interface key validation)
- [x] 6 runtime / proxy tests (session registration, healthcheck-gated startup with hint, kind=remote event payload, proxyTargetAddr resolution, env-host substitution, full HTTP loop)
- [x] 3 e2e goldens (fault rewrite, local-vs-remote parity, mid-run upstream death)
- [x] 2 bundle tests (env.json round-trip + omitempty when unused)
- [x] 2 cmd/faultbox replay tests (warning printer + no-spurious-warn)
- [x] 3 stdlib helper tests (k8s.service / k8s.endpoint / k8s.local)
- [x] 3 documentation gates (spec-language section, connectivity guide, feature-manifest entries)
- [x] `make demo-build` cross-compile clean
- [x] `make demo-container` smoke inside Lima — 4/4 existing tests still pass; confirms the `proxyTargetAddr(svc, iface)` signature change and env-rewrite extension don't regress the seccomp/Postgres/Redis path

## Phasing

| Phase | Commit | Scope |
|---|---|---|
| RFC docs | `46dd7dd` | RFC-036 + RFC-037 drafts |
| Phase 1 | `1a00591` | DSL surface, spec-load validation, fault rule rejection — 32 tests |
| Phase 2 | `a0d963b` | Runtime + proxy + bundle + replay — 8 tests |
| Phase 3 | `9ee1ca9` | k8s discovery helper + spec-language section + connectivity guide + feature manifest + docs gates — 6 tests |
| Phase 4 | `629434d` | E2E fault rewrite, local-vs-remote parity, mid-run-death — 3 tests |

(Commit hashes are pre-rebase; after rebasing onto `aa01812` they shift forward but each phase remains a self-contained commit.)

## Out of scope

- Determinism / record-and-replay — see [RFC-037](https://github.com/faultbox/Faultbox/issues/94).
- Real-pod e2e against a cluster — covered by the `make demo-container` smoke for non-regression; full cluster CI is a follow-up.
- `topic()` / `poll()` event sources on remote services — observe is rejected wholesale in v1; revisit when those sources ship.
- Faultbox-managed Telepresence orchestration — explicit non-goal; user runs `telepresence connect` themselves. Documented in [docs/guides/connectivity.md](docs/guides/connectivity.md).

## Customer-visible changes

- New `service(remote=...)` kwarg + `remotes({...})` typed override + `@faultbox/discovery/k8s.star` helper.
- `.fb` bundles record `remotes: [...]` in `env.json` (omitempty for back-compat).
- `faultbox replay` warns when a bundle declares remotes, points at RFC-037.
- Documentation: new "Remote Services" section in `spec-language.md`, new `docs/guides/connectivity.md`, three new rows in `feature-manifest.md`.

Closes #91 once merged.


avatar29A:
Rebased onto current main (post-[RFC-038](https://github.com/faultbox/Faultbox/issues/95) v0.12.22–v0.12.28). Audited TLS × remote interaction: combination works as-is — `proxyTargetAddr(svc, iface)` from Phase 2 already resolves remote upstreams correctly through the new TLS-aware `preStartProxies` path, and the auto-generated self-signed cert covers `127.0.0.1` so SUT-side TLS verification through the rewritten env addr passes.

New commit `3c03316` adds:
- `TestRemote_TLSUpstream_EndToEnd` — stands up `httptest.NewTLSServer` as the fake remote, exercises full SUT → HTTPS proxy → TLS upstream loop with `interface(..., tls=tls_cert(insecure=True))` + `remote=`.
- New "TLS upstreams (RFC-038 interop)" subsection in [docs/rfcs/0036-remote-services.md](https://github.com/faultbox/Faultbox/blob/epic/rfc-036-remote-services/docs/rfcs/0036-remote-services.md) with worked example, healthcheck nuance for self-signed clusters, per-protocol coverage matrix.
- Mirroring "TLS upstreams (RFC-038)" section in [docs/guides/connectivity.md](https://github.com/faultbox/Faultbox/blob/epic/rfc-036-remote-services/docs/guides/connectivity.md).
- Feature-manifest row updated to cite TLS×remote interop and the increased test count (10 runtime tests).

Full `go test ./...` green; `go vet ./...` clean.

avatar29A:
**Release prep — v0.12.29 (commit 53a3d38).**

- Bumped `cmd/faultbox/main.go` version 0.12.28 → 0.12.29.
- New CHANGELOG entry for 0.12.29 covering the full RFC-036 surface (service(remote=...) + remotes() + k8s.star helper, startRemoteService runtime path, env.json annotation, replay warning, RFC-038 TLS interop) with 49-test coverage breakdown.
- Bottom-of-file link refs updated: `[Unreleased]` now points past 0.12.29; `[0.12.29]` compares release-0.12.28...release-0.12.29.
- RFC-036 status flipped **Draft → Implemented (v0.12.29, 2026-05-02)**. Phasing table marks v1 + v1.1 (TLS interop) shipped.

Release-grade signals:
- `go test ./...` clean (49 new tests; 0 regressions)
- `go vet ./...` clean
- `make demo-build` cross-compile clean
- `make demo-container` Lima smoke 4/4 PASS (from Phase 2, unaffected by version bump)

**Deferred to the user-invoked `/release 0.12.29` skill** (those touch the live Cloudflare deploy and need explicit go-ahead):
- `faultbox-site/src/content/docs/releases.md` — add the 0.12.29 bullet to the "Current" section, demote 0.12.28
- `faultbox-site/src/pages/features.astro` — bump the version table (5-row hard cap; promote 0.12.29 to top, drop oldest, move `current` tag)
- Merge `epic/rfc-036-remote-services` → `main` and tag `release-0.12.29`

Issue #114, opened 2026-05-07
RFC-046: Beyond L1 — gVisor Runtime Roadmap & L5 Research


avatar29A:
# RFC-046: Beyond L1 — gVisor Runtime Roadmap & L5 Research

> **Status: Draft.** Roadmap RFC. Builds on RFC-040 (Determinism Levels) — uses the L0–L5 vocabulary defined there. **Non-binding for v0.13.0.** Captures the post-L1 implementation paths so the v0.14.0+ runtime/level decisions have a documented design baseline.

## Summary

RFC-040 defines the L0–L5 determinism taxonomy and commits v0.13.0 to making L1 a contract. This RFC sketches the implementation paths for *exceeding* L1: Path B (gVisor netstack extraction) and Path C (full Sentry fork) for L2/L3, the L4 Hermetic mode built on Path C, and the L5 instruction-boundary research direction (Hermit / Antithesis class).

None of this work is committed to a specific Faultbox release here. The split from RFC-040 is intentional: RFC-040 is the *contract* for v0.13.0; this RFC is the *map* for what comes next. Customers reading RFC-040 see the L1 promise; engineers reading this RFC see how L2 / L3 / L4 / L5 *could* be reached.

In-tree document: `docs/rfcs/0046-beyond-l1-roadmap.md`.

## Motivation

The motivation chain is the same as RFC-040 — InfoSec primary persona, Antithesis comparison framing, customer demand for stronger reproducibility — but the questions split cleanly:

- **What level do we promise today?** Answered in RFC-040 (L1).
- **How do we reach the levels we don't promise yet?** Answered here.

Without this RFC the post-L1 conversation either bloats RFC-040 (the original 511-line draft) or scatters across notion docs the engineering team can't link to from spec-load error messages. A single in-tree document is the lighter option for both audiences.

## Why the levels split where they do

The progression L3 → L4 → L5 captures three independent dimensions of "more deterministic":

- **L3 → L4 is about *I/O surface*.** L3 reproduces the bytes; L4 forces every I/O into the spec explicitly. Two SUTs at L3 can have wildly different background-I/O profiles; at L4 they cannot — every connection, file open, syscall is named. This is what makes L4 valuable for InfoSec and compliance use cases independent of replay.
[...]

Issue #113, opened 2026-05-04
RFC-044: Spec Language Simplification


avatar29A:
# RFC-044: Spec Language Simplification

> **Status: Draft.** Final RFC of the v0.13.0 epic. Designed alongside RFC-040 / RFC-041 / RFC-042 / RFC-043 to serve as the *design north star* for those implementations, but **implemented last** — after the four preceding RFCs land in v0.13.0-rc1/rc2. Captures cleanups and unifications enabled by the new determinism + non-determinism vocabulary.

## Summary

The four preceding v0.13.0 RFCs (Determinism Levels, Temporal Properties, Exploration Plan, Non-deterministic Operators) introduce a unified conceptual frame: *spec describes non-determinism, engine enumerates all worlds, levels formalize the contract*. With this frame in place, several pre-existing primitives in the spec language become redundant, oversized, or split-where-they-should-be-unified. This RFC consolidates seven cleanups into a single coordinated change so that v0.13.0 ships not just new features but a smaller, clearer language.

The seven items, grouped by impact:

**Major (engine + API):**
1. Withdraw RFC-013 (`param`); fold parameterized scenarios into RFC-043's `choose()`.
2. Refactor the engine's fan-out machinery into one mechanism that handles `fault_matrix`, `choose`, `nondet()`, `interleavings=`, and `probability=` uniformly.
3. Merge `faultbox generate` into `faultbox plan --suggest`.

**Medium (status decision + docs):**
4. Withdraw RFC-002 (`domain()`).
5. Document `remote=` and `reuse=True` as structurally L1-capped features.

**Polish (cosmetic API):**
6. Unify event-source builtins (`stdout`, `stderr`, `topic`, `tail`, `wal_stream`, `poll`) under one `observe.*` namespace.
7. Unify decoder builtins (`json_decoder`, `logfmt_decoder`, `regex_decoder`) under one `decoder("name", ...)` form.

This RFC is **implemented last** in v0.13.0 because items #1–#3 require their replacements (RFC-043's `choose`, RFC-042's plan machinery and `--suggest`) to be shipped first, item #2 is cleanest when all fan-out producers are stable, and the rest naturally land at the end of an epic. The *design* in this RFC, however, ships immediately so RFC-042 and RFC-043 implementers reference it and avoid baking in code that would have to be ripped out.

In-tree document: `docs/rfcs/0044-spec-language-simplification.md`.

### What problem does this solve?

[...]

Issue #112, opened 2026-05-04
RFC-043: Non-deterministic Spec Operators


avatar29A:
# RFC-043: Non-deterministic Spec Operators

> **Status: Draft.** Part of the v0.13.0 epic. Builds on RFC-040 (Determinism Levels) — non-determinism in the spec language only makes sense once we can name what the engine *does* with it. Pairs with RFC-042 (Exploration Plan) which defines the plan-tree machinery this RFC's operators feed into. Subject to RFC-044 (Simplification) — `param()` from RFC-013 will be subsumed by `choose()` defined here.

## Summary

Faultbox's existing fan-out primitives (`fault_matrix`, `param`, `wait_all` interleavings, `probability=` faults) are each specialized expressions of a single underlying concept: **the spec describes non-determinism, and the plan engine enumerates all possible worlds.** This RFC names the concept directly and adds four small Starlark builtins inspired by P-lang and TLA+:

- **`choose([options])`** — non-deterministic finite choice. Plan engine produces one world per option.
- **`nondet()`** — sugar for `choose([True, False])`. Non-deterministic boolean.
- **`halt()`** — terminate the current plan-tree branch. The branch contributes no test execution and no verdict; useful for pruning unreachable or uninteresting paths.
- **`assume(predicate)`** — filter the plan tree. Branches where the predicate is false are pruned.

**Note on naming:** P-lang uses bare `$` as the non-deterministic-boolean operator. Starlark's lexer cannot parse `$` as an identifier or operator — the grammar restricts identifiers to `[A-Za-z_][A-Za-z0-9_]*`. To avoid forking the upstream `go.starlark.net` parser indefinitely, we use `nondet()` as the canonical Starlark-compatible spelling. Functionally identical to P-lang's `$`.

The operators sit alongside existing fan-out primitives, not in replacement of them. `fault_matrix`, `wait_all` interleavings, and `probability=` are syntactic sugar over the same underlying machinery. RFC-044's unified-fan-out refactor consolidates the implementation; this RFC adds the new user-facing primitives.

The conceptual frame: **a Faultbox spec is a partial model of the system's space of behaviors. The plan engine explores that space exhaustively (subject to user-controlled cost knobs from RFC-042). Non-deterministic operators give the spec author the vocabulary P-lang and TLA+ have used for decades to describe model-able systems concisely.**

In-tree document: `docs/rfcs/0043-nondeterministic-operators.md`.

### What problem does this solve?

[...]

Issue #111, opened 2026-05-04
RFC-042: Exploration Plan & Coverage Engine


avatar29A:
# RFC-042: Exploration Plan & Coverage Engine

> **Status: Draft.** Part of the v0.13.0 epic. Builds on RFC-040 (Determinism Levels) — plan visualization is meaningful only because L0 plan determinism guarantees the same tree on every run. Pairs with RFC-041 (Temporal Properties — to be filed).

## Summary

Faultbox specs declare test scenarios via `fault_matrix()`, `fault_assumption()`, parameterized scenarios (RFC-013), and parallel composition (`wait_all`, `wait_first`, `wait_n`). The cross-product of these can produce surprising test counts — a `fault_matrix` with three categories of three options each, parameterized over two retries policies, run inside `wait_all(...)` with three branches, generates 3³ × 2 = 54 plan instances. Today there is no way to see the count *or the structure that produces it* — where forks happen, which compositions multiply with which, where parallel actions diverge — before launching the run. CI cost surprises, coverage gaps, "did fault_matrix do what I expected?" debugging, and "what interleavings could this `wait_all` actually produce?" all bottleneck on the same missing primitive.

This RFC introduces three things:

1. **`faultbox plan <spec>`** — a static analysis that enumerates the plan tree (with forks, parallel actions, and interleaving spaces explicit) without executing anything.
2. **Plan as a first-class bundle artifact.** `faultbox test` runs plan generation as its first phase and writes the resulting tree into the `.fb` bundle. Every test run now ships with its plan attached, so post-hoc analysis (in the report, in MCP queries, in CI dashboards) has the same source of truth as the runtime trace.
3. **A coverage engine** that suggests faults the spec author has not written but probably should — rule-based in v0.13.0, LLM-driven later.

Together they make the test-set visible (the plan tree), the structure visible (forks and parallel actions), the interleaving space visible (what could race with what), and the gaps visible (uncovered dependency edges).

The plan command, bundle integration, and the **report's Plan tab** are v0.13.0 deliverables. The coverage engine is sketched here as direction; the v0.13.0 commitment is foundational rule-based analysis (Section 8). The full LLM-driven coverage engine compounds with RFC-043 (MCP bundle ops, v0.14.0) and lands later. Runtime *interleaving exploration* (DPOR — actually driving the SUT through each possible interleaving) stays out of scope; static *interleaving enumeration* (showing the space) is in scope.

[...]

PR #115, opened 2026-05-10
v0.13.0 — RFC-040 Determinism Levels (L1 contract)

avatar29A:
## Summary

Lands **RFC-040** for v0.13.0 — makes **L1 (mediated-event determinism) a contract**. Spec-level `determinism()` builtin, `unmediated_io` detection across five categories, strict-mode failure (default at L1), CLI override, full docs, RFC drafts, and end-to-end goldens.

Tracking: [#109](https://github.com/faultbox/Faultbox/issues/109). Roadmap companion: [#114 RFC-046](https://github.com/faultbox/Faultbox/issues/114). Sibling RFCs filed but not implemented in this cut: [#110 RFC-041](https://github.com/faultbox/Faultbox/issues/110), [#111 RFC-042](https://github.com/faultbox/Faultbox/issues/111), [#112 RFC-043](https://github.com/faultbox/Faultbox/issues/112), [#113 RFC-044](https://github.com/faultbox/Faultbox/issues/113).

## What ships

**New surface:**
- `determinism(level=, runtime=, strict=, allow=)` top-level builtin. Defaults: `L1` / `default` / `strict=True` / empty allow list. May be called once per spec.
- `service(..., nondeterministic_ok=[...])` per-service tolerance. Unions with spec-level `allow=`.
- Five detection categories — `clock`, `rand`, `dns`, `network-unmediated`, `fs-unmediated` (last is reserved-only).
- `--strict-determinism[=true|false]` and `--no-strict-determinism` CLI overrides on `faultbox test`. Bidirectional and final.
- `outcome="strict_determinism_violation"` flowing through bundle manifest, HTML report, and `Summary.StrictDeterminismViolation`.

**Reserved (parses, errors at spec load):**
- `level="L2"` through `"L5"` and `runtime="gvisor"`. Future migration is non-breaking.

**Visibility / enforcement separation (behaviour change worth flagging):**
- Tolerated categories *still emit* `unmediated_io` events into the trace and bundle. Tolerance only suppresses the strict-mode failure decision — customers see what happened even when they accept it.

## PR sequence (commits on this branch)

| # | Commit | Scope |
|---|--------|-------|
| 1 | `1f5faf4` | `determinism()` builtin + `nondeterministic_ok=` + 24 parse-time tests |
| 2 | `b96a849` | Detection layer (clock/rand/dns/network-unmediated) + 14 tests |
| 3 | `44ad44f` | Strict-mode in `RunTest` + outcome label + 7 strict tests |
| 4 | `d966211` | `--strict-determinism` CLI flag + 3 override tests |
| 5 | `f801652` | Docs (`docs/determinism.md`, spec-language section, README, tutorial Ch24, CHANGELOG) + commits 6 RFC drafts (040/041/042/043/044/046) |
| 6 | `75c6907` | Linux end-to-end testops goldens (5 cases via `poc/leaker`); promoted detection coverage from 🟡 to 🟢 |

## Test plan

- [x] `go vet ./...` clean
- [x] `go test ./...` green host-side (18 packages)
- [x] `go test ./testops/...` green on Lima linux/arm64 (15 active cases including 5 new determinism goldens, 2 skipped pending Docker provisioning)
- [x] Determinism goldens reproduce identically across consecutive runs (twice verified on Lima)
- [x] Determinism unit tests: 48 cases (24 parse-time + 14 detection + 7 strict + 3 CLI override) all green
- [x] `make demo-build` cross-compiles for linux/arm64
- [ ] Reviewer to spot-check: `docs/determinism.md`, [tutorial chapter 24](docs/tutorial/04-safety/24-determinism.md), `docs/spec-language.md` Determinism section
- [ ] Reviewer to validate the RFC-040 §8.3 wording fix against the open-questions resolution log (CLI override is bidirectional, not "promotes only")


avatar29A (commented):
## Principal Engineer Review — RFC-040 Determinism Levels (L1 contract)

Overall: the L1 taxonomy, strict-mode enforcement, per-service escape hatch, and CLI override design are coherent and well-structured. The commit sequence is clean and reviewable. Most blockers below are targeted fixes — no architectural rework required.

---

### 1. Data race on `rt.detAllow` / `rt.detLevel` / `rt.detStrict` — race detector will trip

**`internal/star/determinism.go` — `effectiveAllow`, `strictEffective`**

These helpers read `rt.detAllow`, `rt.detLevel`, and `rt.detStrict` without holding `rt.mu`. They are called from `detectUnmediated` → `makeSyscallCallback` which runs on `handleNotification` goroutines. The writes happen in `builtinDeterminism` during `LoadString`.

In production the sequencing is correct (load completes before goroutines start), but Go's race detector does not reason about application-level sequencing — it requires a synchronisation primitive. `go test -race` will flag this.

**Fix (minimal):** Add a comment on each `det*` field: `// written once at load time before any test goroutine starts; treated as read-only thereafter`. Then either extend `rt.mu` to cover the hot-path reads, or add a sync/atomic field set at end-of-load to satisfy the race detector. Given the existing `rt.mu` pattern the cleanest path is to hold `rt.mu` when reading `detAllow` inside `effectiveAllow`.

---

### 2. `isMediatedAddress` ignores the `ip` parameter — false-mediated verdicts possible

**`internal/star/determinism.go` — `isMediatedAddress`**

```go
func (rt *Runtime) isMediatedAddress(ip string, port int) bool {
    // ip is never used — port-only match
    for _, svc := range rt.services {
        for _, iface := range svc.Interfaces {
            if iface.Port == port || iface.HostPort == port { return true }
        }
    }
    return false
}
```

A SUT connecting to `203.0.113.5:5432` is classified as "mediated" if any declared interface uses port 5432. This is a false negative — the most dangerous kind for a determinism tool. The `TestIsMediatedAddress_DeclaredInterface` test only exercises `127.0.0.1:8080`, so it does not catch this.

**Fix (v0.13.0 acceptable path):** Port-only matching is probably the pragmatic MVP choice; just make the contract explicit:
- Remove `ip string` from the signature (or keep it and add a comment explaining it is unused).
- Add a `// Note: matches by port only; IP is not checked` docstring on the function.
- Update `docs/determinism.md` with a Known Limitations callout: "In v0.13.0, declared interface matching is port-only. A connect() to any host on a declared port is treated as mediated."
- Update the test to exercise a non-localhost IP on the declared port to confirm the behaviour is intentional, not accidental.

---

### 3. `gettimeofday` promised by docs but absent from the detection switch

**`internal/star/determinism.go` (detection) vs `docs/determinism.md`, `docs/rfcs/0040-determinism-levels.md` §3**

The RFC and the user-facing docs both state the `clock` category covers _"clock\_gettime / gettimeofday"_. The detection switch handles only `clock_gettime`. On modern Linux `gettimeofday` is VDSO-accelerated and invisible to seccomp — but on musl, arm64 in certain configs, or older kernels it can reach the real syscall and Faultbox will silently miss it.

**Fix:** Either add `case "gettimeofday"` to the switch (safe, no-op on platforms where VDSO intercepts it), or update every doc that lists "gettimeofday" to explicitly state "gettimeofday is VDSO-accelerated on all supported targets and does not reach seccomp; only clock\_gettime is intercepted." The current state — docs claim coverage, code doesn't deliver it — must be resolved.

---

### 4. Tutorial chapter 24 — incorrect gVisor/VDSO claim

**`docs/tutorial/04-safety/24-determinism.md`**

> "revisit when v0.14.0 lands gVisor (which sees VDSO)"

This conflates two distinct paths from RFC-046:
- **Path B (v0.14.x):** gVisor netstack only — replaces the TCP proxy, does **not** give Faultbox visibility into VDSO.
- **Path C (post-v2.0):** full Sentry fork — processes run inside gVisor Sentry, which intercepts VDSO. This is when VDSO blindness is resolved.

Shipping this as tutorial content will create a concrete user expectation for v0.14.0 that won't be met.

**Fix:**
```
revisit when Path C (Faultbox Sentry fork) lands — see RFC-046.
Path B gVisor (v0.14.x) is network-layer only and does not intercept VDSO.
```

---

### 5. Shipped TODO placeholder in user-facing docs

**`docs/determinism.md` — "Worked examples" section**

The section contains:
> *TODO (v0.13.0 doc work):* before/after Go snippets...

A raw TODO block in a file being merged to `main` will hit readers. Either complete the examples (a service calling `time.Now()`, then fixing it with `nondeterministic_ok`) or delete the section and file a follow-up issue.

---

### 6. RFC-040 escape-hatch table is incomplete — two categories missing

**`docs/rfcs/0040-determinism-levels.md` — "When and how to use `allow`" table**

The table lists `clock`, `dns`, `rand` only. `network-unmediated` and `fs-unmediated` are implemented, appear in the golden fixtures, and are documented in `docs/determinism.md` — but an engineer reading the RFC to understand valid `allow=` values will not find them.

**Fix:** Add rows for `network-unmediated` and `fs-unmediated` matching the descriptions in `docs/determinism.md`.

---

### 7. Implementation plan document — unchecked items + stale §8.3 quote

**`docs/implementation/v0.13.0-rfc-040-plan.md`**

Two issues:

**(a)** The document has unchecked `[ ]` checklist items (including `unmediated_io_caveat` per-service startup event) that were planned but not shipped. Merging a plan document with visibly incomplete boxes into `main` creates ambiguity: is this an oversight or a deferred decision?

**Fix:** Mark deferred items with `[deferred: #<issue>]` pointing to a filed issue, or remove them from the plan.

**(b)** The document quotes the old RFC §8.3 wording ("CLI cannot relax..."). That was the pre-resolution state; the shipped behaviour is bidirectional override (promote and relax). The quote now contradicts the code.

**Fix:** Update or remove the quote. If `docs/implementation/` is internal scaffolding, consider whether it belongs on `main` at all vs. a linked Notion page.

---

**N1. Override tests validate field assignment, not enforcement**

`TestRunAll_AppliesStrictDeterminismOverride{True,False,Nil}` check `rt.strictEffective()` after `RunAll` but do not verify that the override changes a test's pass/fail outcome when `unmediated_io` events are present. A wiring bug in `firstStrictViolation` would pass these tests. Add one behaviour-level test: spec emits a `clock` event, override=false (lenient), assert outcome is `pass`.

**N2. RFC-041 ships with unresolved HTML TODO comments**

`docs/rfcs/0041-temporal-properties.md` has three `<!--TODO: ...-->` blocks. RFC drafts are public on `main`. Track the open questions as issues instead.

**N3. RFC-042 malformed comment**

`docs/rfcs/0042-exploration-plan.md` has `<--TODO: ...>` (not valid HTML comment syntax — `<!--` is required). Will render as broken text in some processors.

**N4. `extractPort` uses `fmt.Sscanf` where `net.SplitHostPort` is idiomatic**

`internal/proxy/proxy.go` — `fmt.Sscanf(addr[idx+1:], "%d", &port)` will parse `"8080abc"` as `8080` silently. `net.SplitHostPort` handles all edge cases including IPv6 brackets and rejects malformed addresses.

**N5. `IsListenPort` uses exclusive lock for a read-only operation**

`internal/proxy/proxy.go` — if `m.mu` is a `sync.RWMutex`, `IsListenPort` should use `RLock` / `RUnlock`. Exclusive lock on a hot-path read adds unnecessary contention.

**N6. `strict_determinism_violation` listed under "Changed" in CHANGELOG, not "Added"**

This is a new outcome value. Listing it under `### Changed` implies it replaced an existing value. Move to `### Added`.

**N7. Rate-limiting risk not documented**

`docs/determinism.md` and RFC-040 don't mention that a service calling `clock_gettime` in a tight loop (common in Go services) will emit one `unmediated_io` event per call, potentially flooding the event log. At minimum, note that `nondeterministic_ok=["clock"]` avoids log saturation for high-frequency callers, and flag the deferral of rate-limiting to a future RFC.

**N8. `poc/leaker` DNS case uses live external resolver**

`poc/leaker/main.go` — `net.DialTimeout("udp", "8.8.8.8:53", ...)` requires outbound UDP/53 in CI. In locked-down CI environments this will fail with a connection error rather than a timeout, potentially breaking the golden. Consider `127.0.0.1:53` or make the target configurable via env var.

**N9. `NormalizeTrace` comment slightly misleading**

The comment says `detail` carries a destination address that varies due to "ephemeral ports" — but `detail` carries the *destination* `8.8.8.8:53`, which is stable. The *source* port is ephemeral. Update the comment.

---

## QUESTIONS — need answer before merge

**Q1. What happens when a fault rule *denies* a connect to an unmediated address?**

`detectUnmediated` fires regardless of `evt.Decision`. A fault rule that denies a connect to `203.0.113.5:5432` will still emit `unmediated_io[network-unmediated]` and trigger strict-mode failure — even though the connect was already blocked. Is this intended? If the SUT is fault-injected and the connect never succeeds, is it still a determinism violation? If not, add a guard: `if evt.Decision != "allow" { return }` in `detectUnmediated` for connect events.

**Q2. `RegisterListenAddr` ordering vs. first SUT `connect` syscall**

`IsListenPort` reads `m.listeners`. Is `RegisterListenAddr` guaranteed to be called before any SUT `connect` syscall can be dispatched? If proxy startup is async relative to seccomp filter installation, there is a window where `IsListenPort` returns false for a port that will become mediated. A comment confirming the startup ordering invariant (or a test) would close this.

**Q3. Strict-mode violation vs. test body error priority**

If both a Go-panic/setup error AND `unmediated_io` events exist for the same test, does the `strict_determinism_violation` outcome shadow the underlying `error` outcome? Is this the intended priority? Documenting the ordering (even in a code comment) would help future engineers.

---

## Minor

- `fmt.Sprintf("%d", evt.PID)` where PID is `uint32` — `strconv.FormatUint(uint64(evt.PID), 10)` avoids an allocation in the hot callback path.
- The `len(args) != 0` guard before `UnpackArgs` in `builtinDeterminism` is load-bearing (prevents positional args being silently mapped to `level`). Add a comment so it doesn't look like redundant defensive code.
- Tutorial chapter numbering has a gap at 21–23. Intentional? Worth a grep to confirm no cross-references use sequential numbering.

---

## Summary

| Severity | Count |
|----------|-------|
| Blocking | 7 |
| Non-blocking | 9 |
| Questions | 3 |

All blockers are targeted fixes. The biggest one (isMediatedAddress IP, gettimeofday coverage, gVisor/VDSO tutorial claim) need correction before users build mental models from incorrect docs. The race on `det*` fields needs resolution before `go test -race` is added to CI (which it should be for a determinism PR). Happy to re-review once the blocking items are addressed.

avatar29A (commented):
## Follow-up Review — commit `60f2d394`

Verified the fix commit against all 7 blocking issues, 5 non-blocking items, and 3 open questions. Most are resolved correctly. Three issues remain that need attention before merge.

---

### Remaining concerns

**[Still blocking] B3 — `gettimeofday` case is dead code at runtime**

The `case "clock_gettime", "gettimeofday":` branch in `detectUnmediated` is unreachable at runtime because `requiredSyscallsForService` installs the seccomp filter for `clock_gettime` but not for `gettimeofday`. The callback is never invoked for `gettimeofday`, so the case never fires. Meanwhile `docs/determinism.md` §Known Limitations now claims:

> "The `clock_gettime` raw syscall path IS caught"

...which implies (correctly) that `clock_gettime` is intercepted, but the RFC-040 table and some surrounding text reference both syscalls as if both are caught. The fix made the documentation more prominent about `gettimeofday` without wiring it up, which leaves the implementation and docs misaligned in a new way.

**Fix options (pick one):**
- Add `"gettimeofday"` to `requiredSyscallsForService` alongside `clock_gettime`. Safe — VDSO intercepts it on all supported targets so it's a no-op in practice; catches the fallback path on musl/arm64/old kernels as documented.
- Remove `"gettimeofday"` from the detection switch case and every doc reference that implies it is intercepted. State clearly: "only `clock_gettime` is in the seccomp filter; `gettimeofday` is not intercepted even on the raw-syscall path."

The current state — case in switch, absent from filter — is worse than either original state.

---

**[Incomplete fix] B1 — `requiredSyscallsForService` reads `rt.detLevel` without `rt.mu`**

The B1 fix correctly adds `rt.mu` guards to `effectiveAllow`, `strictEffective`, and `detectUnmediated`. But `requiredSyscallsForService` reads `rt.detLevel` at its first line without holding `rt.mu`. This function is called from `RunTest` after `LoadString` completes, so it is architecturally safe — but it is not aligned with the locking discipline established for the other `detLevel` reads, and a `go test -race` run that introduces any scheduling interleaving between this path and `builtinDeterminism` could still trip. The fix is one line: acquire `rt.mu` around the `rt.detLevel` read at the top of `requiredSyscallsForService`.

---

**[Minor, no action required] effectiveAllow doc comment is misleading post-fix**

The new comment on `effectiveAllow` says `rt.mu` is held "for the entire read" but `svc.NondeterministicOK` is iterated after the unlock. Not a real race (it is write-once during `LoadString`), but the comment is inaccurate to whoever reads it next. Suggest: "Holds rt.mu for detAllow; NondeterministicOK is write-once at load time and needs no lock."

---

### What verified correctly — no further action

| Issue | Status |
|-------|--------|
| B1 (`effectiveAllow`, `strictEffective`, `detectUnmediated` mutex) | Fixed correctly |
| B2 (`isMediatedAddress` doc comment + Known Limitations table) | Fixed correctly |
| B4 (tutorial ch24 — Path C / RFC-046 reference) | Fixed correctly |
| B5 (TODO placeholder replaced with tutorial pointer) | Fixed correctly |
| B6 (RFC-040 table — `network-unmediated` + `fs-unmediated` rows added) | Fixed correctly |
| B7 (implementation plan — unchecked items and stale §8.3 quote) | Fixed correctly |
| N4 (`extractPort` → `net.SplitHostPort` + `strconv.Atoi`, IPv6 tested) | Fixed correctly |
| N8 (`poc/leaker` DNS_RESOLVER env var with default) | Fixed correctly |
| Q1 (denied-connect rationale comment) | Documented correctly |
| Q2 (`RegisterListenAddr` ordering invariant comment on `IsListenPort`) | Documented correctly |
| Q3 (priority ordering comment in `RunTest`) | Documented correctly |
| N2/N3 (RFC-041 TODOs cleaned, RFC-042 malformed tag fixed) | Fixed (cannot diff against prior version, but new content is clean) |

---

### One observation on N1

`TestStrictOverrideFalse_SuppressesEnforcement` has a dead code path — the `if rt.strictEffective() { ... firstStrictViolation ... }` block at the end is unreachable because `strictEffective()` returns false under the test's conditions. The trace-presence check and the `strictEffective()` return assertion are meaningful; the inner `firstStrictViolation` call is not exercised. Not blocking, but worth cleaning up for clarity.

---

### Conclusion

Fix the `gettimeofday` filter gap (B3) and the `requiredSyscallsForService` lock gap (B1 remainder). Both are single-line changes. Once those are in, the PR is ready to merge.

avatar29A (commented):
LGTM. All blocking issues resolved — the two remaining blockers from the second round (gettimeofday absent from seccomp filter, unguarded detLevel read in requiredSyscallsForService) are correctly fixed in f87777e5. Changes are minimal and targeted. Approve to merge.

Issue #76, opened 2026-04-24
Panic-safe bundle flush: write partial .fb before crashing


avatar29A:
## Summary

Today when Faultbox itself panics mid-suite, the `.fb` bundle is lost because the process dies before `emitBundle` runs. Reported by inDrive Freight (v0.11.1 feedback, P1): "the panic-on-fault-apply loses the bundle — partial bundle with fewer tests is more useful than none."

## Scope

- Install a top-level `defer recover()` in `cmd/faultbox/main.go:testStarCmd` (and `replay`, etc.) that:
  1. Captures stack trace into `manifest.json` as a new field `crash: {panic, stack, last_test}`.
  2. Flushes whatever `SuiteResult.Tests` and event log we've accumulated.
  3. Writes the bundle with the usual filename but marks the manifest `schema_version_note: "partial"` (so tools know not to trust completeness).
  4. Re-raises the panic so CI still fails red.
- Add crash-safety test: a deliberate panic inside `makeFaultScenarioRunner` must still produce a valid openable `.fb`.

## Non-goals

- Not a general error-recovery mechanism. Specs with `assert_*` failures still produce normal bundles; this is purely for *runtime* panics.

Customer reference: inDrive Freight v0.11.1 report, ask #3.

Issue #84, opened 2026-04-24
CI coverage gate: internal/proxy/*.go must have *_test.go with passthrough test


avatar29A:
## Summary

Add a pre-release / CI check that fails if any `internal/proxy/*.go` source file (excluding `_test.go` and generated files) is missing a sibling `*_test.go` that includes at least a **passthrough round-trip test**. This directly addresses the process gap that let the v0.11.1 gRPC corruption ship to inDrive Freight.

## Motivation

**Process gap discovered during v0.11.2 hotfix** (2026-04-25):

- `internal/proxy/grpc.go` had **zero tests**. The raw-bytes codec bug that caused Freight's `message is *[]uint8, want proto.Message` would have been caught by the most basic byte-identity round-trip test.
- v0.11.2 added `internal/proxy/grpc_test.go` with that exact regression test — but only after the customer incident.
- No structural safeguard prevents another `internal/proxy/*.go` plugin from shipping untested in the future.

The proxy layer is on the SUT data path. A broken proxy silently corrupts customer traffic even when no fault is injected — the worst failure mode for a fault-injection tool (we degrade correctness instead of testing it).

## Proposed Check

A lint/CI step that:
1. Walks `internal/proxy/` for `*.go` files, excluding `_test.go` and anything under `sqlmatch/` (already well-covered).
2. For each, requires a sibling `<name>_test.go` exists.
3. Greps each `*_test.go` for a passthrough test — pattern: test function name contains `Passthrough`, `RoundTrip`, or `DoesNotCorrupt`, AND the test exercises a real server↔proxy↔upstream byte-identity round-trip (heuristic: presence of both a listener + a client invocation in the same test).

Implement as a `make check-proxy-coverage` target, wire into the `release` skill's pre-release checklist (step 1: "Full Regression Testing") and into GitHub Actions CI.

## Scope

- New file: `scripts/check-proxy-coverage.sh` (or Go program at `cmd/tools/proxycov`)
- Update: `Makefile` — add `check-proxy-coverage` target, include in `make lint` and `make test`
- Update: `.github/workflows/ci.yml` — run `make check-proxy-coverage` as a required step
- Update: `.claude/skills/release/SKILL.md` — add coverage check to step 1
- Update: `CLAUDE.md` — note the invariant under "Code Standards"

## Non-goals

- General per-package coverage percentage gates (separate concern; can get noisy).
[...]

PR #116, opened 2026-05-11
feat: RFC-041 Temporal Properties (v0.13.0)

avatar29A:
## Summary

Ships RFC-041 (Temporal Properties & Monitors) on top of the v0.13.0
determinism contract. Adds five composable temporal primitives, a
declarative \`test()\` builtin, a three-valued PASS/FAIL/INCONCLUSIVE
verdict, and the trace API + event-log indexes that back the predicate
language.

User guide: \`docs/temporal.md\`. RFC: \`docs/rfcs/0041-temporal-properties.md\`.

## PR sequence (7 commits)

| # | Commit | Scope |
|---|--------|-------|
| 1 | \`a2a0148\` | Trace API + event log secondary indexes + \`match\` builtin (§8.5) |
| 2 | \`74d69f8\` | Monitor sandbox: denylist validator + thread tagging (§5.4, §8.7) |
| 3 | \`7929eed\` | \`monitor()\` rewrite — \`on=\` matcher + state machine (§5.4) |
| 4 | \`3489120\` | \`eventually()\` + \`always()\` expectations + Verdict type (§5.1, §5.2) |
| 5 | \`2c914ec\` | \`await_stable()\` + \`await_event()\` body-blocking primitives (§5.3) |
| 6 | \`eb96074\` | PASS/FAIL/INCONCLUSIVE lifecycle + \`test()\` builtin (§5.5, §8.6, §8.8) |
| 7 | \`aa35d4a\` | Docs — temporal.md, spec-language, manifest, RFC status (§8.10) |

Recommended merge: squash or fast-forward — every commit is green on its own.

## What ships

**Five new primitives** (RFC-041 §5):
- \`eventually(p, anchor=)\` — liveness; final eval at Termination per §5.5
- \`always(p, between=)\` — invariant; immediate fail on first violation
- \`await_event(matcher_or_predicate)\` — block body until match; returns the event
- \`await_stable(quiescence_window=, ignore=)\` — block until no non-ignored event
- \`monitor(name, on=, state_init=, update=, check=)\` — state-machine monitor with per-test memory

**Test lifecycle:**
- New \`test(name, body=, setup=, expect=, timeout=, terminate_when=, clock=)\` declarative builtin
- Three-valued verdict: \`pass\` / \`fail\` / \`error\` / \`inconclusive\`
- Per-test wall-clock timeout via \`context.WithTimeout\`
- \`terminate_when\` watching cancels the body context when its predicate holds
- Body now runs in a goroutine; select on body-done / terminate_when / timeout determines the §5.5 Termination cause
- Panic recovery preserved (issue #76): panic → \`Result=\"error\"\` → \`SuiteResult.Crash\`
- CLI exit codes: 0 pass, 2 any-fail, **3 inconclusive-only**

**Foundations:**
- \`TraceVal\` with \`trace.event/events/first/last/count/events_between/events_within/causal_chain\`
- \`EventVal\` with causal operators (\`happens_before/after\`, \`concurrent_with\`, \`preceded_by/within\`, \`followed_by/within\`, \`directly_caused_by\`, \`duration_since\`, \`same_service_as\`, \`same_correlation_as\`)
- \`EventListVal\` sequence with \`.map/.filter/.reduce\`
- \`match.event/any/all/never\` namespace
- Secondary indexes on \`EventLog\` (by type, by service) built incrementally in \`Emit()\`
- Monitor sandbox: spec-load static check rejecting fault/service/assert_*/parallel/eventually/await_*/recursive monitor calls inside \`update\`/\`check\` lambdas

**§8.8 reserved syntax:** \`clock=\"virtual\"\` on \`test()\`, \`await_stable()\`, and \`await_event()\` parses but errors with \`\"requires gVisor (Path C); not available in this release\"\` so the L1→L3 migration is a substrate swap, not a spec rewrite.

## Compatibility

- Legacy \`def test_*()\` continues to work — no per-test timeout, no \`terminate_when\`. Inline \`eventually(...)\` calls are still Finalized at body return.
- Old \`monitor(callback, service=, ...)\` signature is **removed** outright (per \`feedback_no_real_users.md\`: no users, no compat shim). All in-tree tests updated.
- Pre-RFC-041 specs serialize identically — the new \`Inconclusive\` field on \`SuiteResult\`/\`TraceOutput\` is omitted when zero.

## Test plan

- [x] \`go test ./...\` green (cmd, docs, all \`internal/*\`, testops)
- [x] \`go vet ./...\` clean
- [x] New test files:
  - \`trace_test.go\` (lookup operators, secondary indexes, causal relations)
  - \`match_test.go\` (matcher composition + edge cases)
  - \`monitor_sandbox_test.go\` (denylist coverage + AST walker)
  - \`temporal_test.go\` (Expectation lifecycle, finalize-cause matrix, anchors)
  - \`monitor_test.go\` (state machine, on= filter, per-registration state, sandbox enforcement)
  - \`await_test.go\` (quiescence, timer reset, ignore filter, eager check, ctx cancellation, §8.8 reservations)
  - \`lifecycle_test.go\` (test() discovery, per-test timeout → INCONCLUSIVE, eventually unsatisfied → FAIL, panic → error, legacy def test_*() compatibility, Inconclusive counter)
- [ ] **Testops goldens** for end-to-end primitives via the leaker binary — deferred follow-up; unit + lifecycle suite already locks down semantics.

## Known follow-ups

- 8.9 testops goldens (Linux corpus specs exercising the primitives end-to-end via \`/tmp/faultbox-leaker\`)
- Tutorial chapter \`docs/tutorial/04-safety/25-temporal-properties.md\` — light material now lives in \`docs/temporal.md\`; a dedicated walkthrough can land separately
- CLI exit code \`3\` choice documented in \`main.go\` — diverges from RFC's literal fail=1/inconclusive=2 to keep our existing fail=2 stable for current CI integrations

## Sections shipped

| §   | Status |
|-----|--------|
| 8.1 \`eventually\` | ✅ |
| 8.2 \`always\` | ✅ |
| 8.3 \`await_stable\`/\`await_event\` | ✅ |
| 8.4 \`monitor\` + sandbox | ✅ |
| 8.5 Trace API + indexes | ✅ |
| 8.6 Lifecycle + \`test()\` | ✅ |
| 8.7 Spec-load checks | ✅ |
| 8.8 Reserved kwargs | ✅ |
| 8.9 Unit tests | ✅ (goldens = follow-up) |
| 8.10 Docs | ✅ |


avatar29A (commented):
## Principal Engineer Review — RFC-041 Temporal Properties & Monitors

Overall shape is good: the three-valued verdict, ExpectationVal interface, TestConfig/lifecycle separation, monitor state machine, and sandbox AST walker are well-conceived. The commit sequence is readable and each step builds sensibly on the previous. However there are several **blocking correctness and API-contract bugs** that must be resolved before merge.

---

### 1. Deadlock: `Emit()` holds write lock while subscriber callbacks re-enter the log

**`internal/star/events.go` — `Emit()`**

`Emit()` holds `l.mu.Lock()` through its entire execution, including while calling subscriber callbacks (the `defer l.mu.Unlock()` at the top of the function). The new PR adds three subscriber callbacks that call back into the EventLog:

- **`terminate_when` subscriber** (`runtime.go`): calls `testCfg.TerminateWhen.Evaluate(twThread, rt.events)` → `starlark.Call(thread, predicate, Tuple{&TraceVal{log}})` → any predicate method (`trace.event(...)`, `trace.events(...)`, `trace.last(...)`) → `l.mu.RLock()` → **deadlock**.
- **`RegisterMonitor` subscriber** (`runtime.go`): calls `starlark.Call(thread, m.Check, Tuple{evVal, state})` where `evVal` stores a pointer to `rt.events`. Any monitor `check=` predicate that calls `event.preceded_by()`, `event.followed_by()`, `event.directly_caused_by()`, `event.duration_since()`, or `event.same_service_as()` → these all call `l.mu.RLock()` → **deadlock**.
- **`AlwaysExpectation.Evaluate`** if ever called from a per-event subscriber: same path.

`sync.RWMutex` is not re-entrant. The process deadlocks silently on any spec that uses `terminate_when=`, `monitor(check=...)` with causal predicates, or any future per-event evaluator.

**Fix:** Release `l.mu` before dispatching subscribers. Replace `defer l.mu.Unlock()` with an explicit unlock before the subscriber dispatch loop. Keep a separate `l.subMu` (or snapshot-under-lock + dispatch-unlocked) to protect the subscriber list.

---

### 2. `always()` "immediate failure" promise is not delivered — evaluated only at Finalize

**`internal/star/temporal.go`, `internal/star/runtime.go`**

The RFC §5.2, `docs/temporal.md`, `docs/spec-language.md`, and the `ExpectationVal` interface docstring all state `always(p)` "fails immediately on the first violation." This is false. Neither `always()` nor `eventually()` has an event-log subscription. Both are evaluated only when `Finalize()` is called (at body-return, timeout, or terminate_when). A body using `await_stable()` can run to the full `timeout=` budget even after `always()` has been violated.

Consequence: a test with `expect = always(lambda t: not has_double_spend(t))` will not terminate when the invariant breaks mid-run. The final verdict will be FAIL, but the timing contract is wrong.

**Fix:** Add a per-event subscriber for each registered `AlwaysExpectation` that calls `Evaluate()` and cancels `bodyCtx` on `VerdictFail`. This requires issue #1 to be fixed first (lock ordering).

---

### 3. Data race on `rt.inTest`

**`internal/star/runtime.go` — `RunTest`; `internal/star/builtins.go` — `builtinMonitor`**

`rt.inTest` is a bare `bool` field written by the main goroutine (`rt.inTest = true` before body starts, `rt.inTest = false` after the select) and read by the body goroutine via `builtinMonitor`. When `waitBodyExit` times out (issue #4 below), the main goroutine sets `rt.inTest = false` while the leaked body goroutine may still be reading `rt.inTest`. `go test -race` will flag this.

**Fix:** Replace with `atomic.Bool`.

---

### 4. `vcHappensBefore` is incorrect for asymmetric vector-clock key sets

**`internal/star/trace.go` — `vcHappensBefore()`**

The function only iterates `vc1`'s keys:
```go
for k, v1 := range vc1 {
    v2 := vc2[k]
    if v1 > v2 { return false }
    if v1 < v2 { strictLess = true }
}
return strictLess
```

For `vc1 = {"a": 1}`, `vc2 = {"a": 1, "b": 5}`: the loop checks only key `a` (equal), finds no strict less, returns `false`. Correct answer is `true` — vc2 has progressed further on dimension `b` that vc1 has no knowledge of (vc1["b"] = 0 < 5).

This affects `trace.causal_chain()`, `event.happens_before()`, `event.preceded_by()`, `event.followed_by()`, `event.directly_caused_by()` — the entire causal API gives wrong results in multi-service traces where services have asymmetric VC knowledge. The test suite covers only symmetric same-key cases.

**Fix:** Iterate the union of keys from both clocks (or iterate vc2's keys in a second pass):
```go
func vcHappensBefore(vc1, vc2 map[string]int64) bool {
    strictLess := false
    for k, v2 := range vc2 {
        v1 := vc1[k]
        if v1 > v2 { return false }
        if v1 < v2 { strictLess = true }
    }
    for k, v1 := range vc1 {
        if _, seen := vc2[k]; seen { continue }
        if v1 > 0 { return false }
    }
    return strictLess
}
```

---

### 5. `trace.event()` documented as returning the *first* match — implementation returns *last*

**`docs/spec-language.md` line ~457; `docs/temporal.md` Returns column**

`docs/spec-language.md`: "`trace.event(type=..., **fields)` → first matching `EventVal` or None"
`docs/temporal.md` Returns column: "First matching event (or `None`)"

Both are wrong. The implementation, code comments, and `trace_test.go` all confirm `trace.event()` returns the **most recent** (last-by-emission-order) matching event. `trace.last` is defined as equivalent to `trace.event`. The `trace.first` accessor is what returns the first match.

A user following the docs and using `trace.event()` to find the "first" request event will silently get the last one.

**Fix:** Correct both doc locations to "most recent (last) matching event." Add a note: "use `trace.first(...)` for the earliest."

---

### 6. `duration_since()` doc example uses string comparison — silently produces wrong results

**`docs/temporal.md`, `docs/spec-language.md` examples**

The docs show:
```python
t.event(type="api.response").duration_since(t.event(type="api.request")) < "200ms"
```

`event.durationSince()` returns `d.String()` — a Go duration string like `"150ms"` or `"1.5s"`. Starlark compares strings lexicographically: `"500ms" < "200ms"` is `False` (correct by accident), `"1.5s" < "200ms"` is `True` ("1" < "2") — wrong. Any duration crossing a magnitude boundary gives a wrong result.

**Fix (two valid options):**
- Change `durationSince` to return nanoseconds as a Starlark int (most composable, idiomatic), update docs to use numeric comparison.
- Or document that `duration_since()` returns a string and provide a working comparison helper (`duration_ms(e)`) that parses it. Either way the current doc example is broken.

---

### 7. `inconclusive` tests silently mapped to `"passed"` in bundle manifest and HTML report

**`cmd/faultbox/main.go` — `testRowsFromResult`; `internal/bundle/`**

The `switch` on `tr.Result` in `testRowsFromResult` has no case for `"inconclusive"`, falling through to `default` which produces `outcome = "passed"`. INCONCLUSIVE tests appear as green in the HTML report and as `Passed` in the bundle manifest.

The PR's feature list says "Inconclusive distinguished from Fail in SuiteResult, TraceOutput, and CLI exit code." The bundle manifest and HTML report are excluded from that list — but without them, CI dashboards built on the bundle artifact will misread INCONCLUSIVE as a pass. A follow-up issue should be filed (or the outcome propagated here).

---

### 8. `always(between=("body_start", "stable"))` — string lifecycle anchors are accepted but do nothing

**`internal/star/temporal.go` — `AlwaysExpectation.Evaluate`; `internal/star/runtime.go`**

The docs in both `docs/temporal.md` and `docs/spec-language.md` present `between=("body_start", "stable")` as a working feature. At runtime:
- `betweenStart` string anchor: `windowStarted` defaults to `true` on the first `Evaluate()` call regardless. The comment says "SetWindowStarted once PR 6 lands" — but `SetWindowStarted` is not defined anywhere in this diff or the existing codebase. The window opens immediately.
- `betweenEnd` string anchor: `a.betweenEnd.matcher == nil` for string anchors, so the window close check is skipped and the window never closes.

String anchors are parsed and silently ignored. The docs must either be corrected to show only matcher-form anchors work, or `SetWindowStarted/SetWindowEnded` must be implemented and wired into the test lifecycle before merge.

---

**N1. Shared `starlark.Thread` reused across subscriber invocations in `terminate_when`**

Once the Emit lock is released before dispatch (fix #1), concurrent `Emit()` calls from separate goroutines could invoke the `terminate_when` subscriber concurrently, reusing the same `twThread`. `starlark.Thread` is not goroutine-safe. Fix: create a fresh thread per subscriber invocation.

**N2. Goroutine leak when body hangs past `waitBodyExit` deadline**

If `waitBodyExit(bodyDone, 2s)` expires (body blocked in non-cancellable code), the leaked body goroutine continues running against `rt` while the next test's `RunTest` resets `rt` fields. The `defer rt.setTestContext(nil)` mitigates but does not fully close the window. Separately: the 2-second timeout is hardcoded and compounds in fault matrices (20 rows × 2s = 40s extra). Consider exposing via `RunConfig`.

**N3. `terminate_when=` accepts `always()` — semantically backward**

`always(p)` returns `VerdictFail` on violation, `VerdictPending` otherwise. Using it as `terminate_when` means "terminate when the invariant is violated." That is a valid use case but is undocumented and the error message says "must be `eventually(...)` or `always(...)`" without explaining the `always` semantics. Consider restricting to `eventually` only, or documenting the invariant-violation trigger explicitly.

**N4. `match.event()` silently drops non-string kwarg values**

`starlark.AsString` returns `("", false)` for non-string values; the `false` is not checked. `match.event(type=42)` silently becomes a matcher for `type == ""`. Return an error for non-string field values.

**N5. `events.sum()` is in `AttrNames()` but always errors**

`EventListVal.AttrNames()` advertises `"sum"`. Every call to `sum()` returns an error. Remove from `AttrNames()` or implement it before the API is published.

**N6. `test()` silently overwrites a duplicate test name**

`rt.testConfigs[fullName] = cfg` overwrites with no error. Return an error if the name is already registered.

**N7. `trace.events_within(of=None)` broken**

When `of=None` (the default), the anchor time is Go's zero `time.Time` (year 1). Every real event's timestamp will be far outside any reasonable `window` from year 1. The function should treat `of=None` as "no time filter, return all matching events."

**N8. Wrong exit code in two `runtime.go` comments**

`SuiteResult.Inconclusive` field comment and the `RunAll` "inconclusive" case comment both say "CLI exit code 2." Correct value is **3**. Exit code 2 is for `Fail > 0`.

**N9. RFC §8.6 exit-code divergence not annotated in the RFC document**

`main.go` correctly documents the deviation (RFC says 1 for fail; implementation uses 2 to preserve existing CI). The RFC itself is not updated. A reader of the RFC will see conflicting information.

**N10. Monitor sandbox denylist docs are missing four entries**

`monitor_sandbox.go` denylists `"trace"`, `"trace_start"`, `"trace_stop"`, and `"events"`. Neither `docs/temporal.md` nor `docs/spec-language.md` mentions them in the sandbox restriction table.

**N11. `def`-form sandbox bypass has no filed issue**

`TestValidateMonitorLambdas_DefByReferenceIsKnownLimitation` documents the bypass correctly. File a GitHub issue so the limitation doesn't silently disappear from the backlog.

---

## QUESTIONS — need answer before merge

**Q1. Is the `always()` per-event evaluation gap intentional or unintentionally omitted?**

If intentional (deferred to avoid calling Starlark predicates from the hot Emit path), then the docs must be corrected everywhere they say "fails immediately" to say "evaluated at Termination." If unintentionally omitted, fix before merge — it is the defining semantic of `always()`.

**Q2. Is `inconclusive → "passed"` in the bundle intentional?**

CI dashboards built on the bundle manifest will misread inconclusive as pass. Should `bundle.TestRow` get an `"inconclusive"` outcome value? If deferred, file a follow-up issue.

**Q3. Is `duration_since()` intended to return a string or an integer?**

The causal API is new. Now is the cheapest time to change the return type. If it returns nanoseconds as an int, comparison operators work naturally. If it stays a string, every user who writes `< "200ms"` gets silently wrong behavior.

**Q4. Are testops goldens a merge gate?**

The PR defers end-to-end goldens. Given that `always()` immediate-failure semantics and causal-chain correctness have no integration test coverage, the unit tests are not sufficient to catch semantic regressions. Is there a timeline for the goldens, and is there a blocker on merging without them per the #84 coverage gate?

---

## Summary

| Severity | Count |
|----------|-------|
| Blocking | 8 |
| Non-blocking | 11 |
| Questions | 4 |

The deadlock (#1) is the highest priority — it renders `terminate_when` and any monitor `check=` with causal predicates non-functional. The `always()` immediate-failure gap (#2) is the most visible semantic correctness issue. The `vcHappensBefore` bug (#4) silently corrupts the causal-chain API for any multi-service trace. All three must be fixed before this ships as a v0.13.0 feature.

avatar29A:
Thanks for the thorough review. Addressed in 49ea8e0:

## Blocking (all 8 fixed)

1. **Emit() deadlock** — `events.go` now releases `l.mu` before dispatching subscribers. Subscriber callbacks read the log through `TraceVal` methods that acquire `l.mu.RLock()` themselves; the new event is already visible at that point.
2. **always() per-event eval** — `runtime.go` installs an `alwaysFired` watcher subscriber that walks `snapshotExpectations()` after every event, calls `AlwaysExpectation.Evaluate`, and cancels `bodyCtx` on the first `VerdictFail`. New select arm classifies as `TerminationImmediateFail`.
3. **`rt.inTest` race** — promoted to `atomic.Bool` (Load/Store call sites updated in `runtime.go` + `builtins.go`).
4. **`vcHappensBefore`** — rewritten to iterate the union of both clocks' key sets. Missing keys are treated as zero. Added test cases for asymmetric subset/superset and the both-empty / empty-vc1 edges.
5. **`trace.event()` docs** — corrected in `docs/temporal.md` and `docs/spec-language.md` to "most recent (last) matching event" (alias for `trace.last`); user pointed at `trace.first()` for the earliest.
6. **`duration_since` return type** — now returns `int64` nanoseconds (not a string). Added `duration("200ms")` helper builtin that also returns ns so the two compose with `<`, `<=`, etc. Doc example fixed.
7. **Inconclusive in bundle/report** — `cmd/faultbox/main.go` `testRowsFromResult` adds a case for `"inconclusive"`; `bundle.Summary` gained an `Inconclusive` counter; HTML report (`app.js`) handles the outcome with the `warn` class and the header pill surfaces the count.
8. **Named lifecycle anchors** — `AlwaysExpectation.SetWindowStarted/SetWindowEnded` implemented; `runtime.signalLifecycleAnchor()` fires `body_start` at body entry, `body_end` at body exit, and `stable` after `await_stable` returns. String anchors now actually do something.

## Non-blocking (most fixed)

- **N1** — fresh `starlark.Thread` per subscriber invocation in both `terminate_when` and the new always-watcher.
- **N4** — `match.event()` and `argsToMatcher` reject non-string kwarg values explicitly.
- **N5** — `events.sum()` removed from `AttrNames()` and the stub deleted.
- **N6** — `test()` returns an error on duplicate names.
- **N7** — `trace.events_within(of=None)` no longer filters against the zero timestamp.
- **N8** — corrected the two "CLI exit code 2" comments in `runtime.go` to reference exit code 3.
- **N9** — RFC §8.6 annotates the exit-code rebase (RFC reads FAIL=2 / INCONCLUSIVE=3 now, with a note explaining the preservation of Faultbox's pre-RFC-041 FAIL=2 CI convention).
- **N10** — monitor sandbox docs in `docs/temporal.md` + `docs/spec-language.md` now list `trace`, `trace_start`, `trace_stop`, `events` alongside the other denied builtins.

## Deferred / open

- **N2** (waitBodyExit goroutine leak) — kept the 2s deadline. Filing as follow-up; matrix amplification is real but the safer fix needs a body-cancellation contract.
- **N3** (`terminate_when=always(...)` semantics) — left the surface as-is for now; can tighten the error message in a follow-up.
- **N11** (def-form sandbox bypass) — follow-up issue.
- **Q4** (testops goldens for temporal primitives) — still deferred; added to the PR body's "known follow-ups" but not blocking per the v0.13.0 plan.

Tests: `go test -race ./...` clean.

avatar29A (commented):
## Follow-up Review — commit `49ea8e01`

Verified the fix commit against all 8 blocking issues and all non-blocking items. Everything is addressed correctly. One minor cleanup note before LGTM.

---

### One cleanup item (no action required to merge)

**`alwaysFailExp`, `alwaysFailMsg`, `alwaysFailMu` are dead code — `runtime.go`**

In the always-watcher subscriber, the three variables are captured, stored under `alwaysFailMu`, then explicitly suppressed with `_ = alwaysFailExp; _ = alwaysFailMsg`. They are never read. The violation is correctly surfaced through the standard path: `Evaluate` sets `a.violated = true` and `a.violationMsg` as a side effect, then `Finalize` picks it up. The comment above the `_` lines ("select arm consults it") is inaccurate — the select arm never reads them.

Not blocking; correctness is unaffected since `Finalize` handles it. Clean up in a follow-up by removing the three variables and the mutex, and correcting the comment.

---

### What verified correctly — all items

| Issue | Status |
|-------|--------|
| B1 — `l.mu` released before subscriber dispatch (deadlock fix) | Verified correct |
| B2 — `always()` per-event watcher with fresh thread, bodyCancel, Once | Verified correct |
| B3 — `rt.inTest` → `atomic.Bool` | Verified correct |
| B4 — `vcHappensBefore` two-pass union-of-keys iteration + new tests | Verified correct |
| B5 — `trace.event()` docs corrected to "most recent (last)" in both files | Verified correct |
| B6 — `duration_since()` returns `int64` ns; `duration()` helper added and registered | Verified correct |
| B7 — `inconclusive` in `testRowsFromResult`, `bundle.Summary`, `summaryFromTests`, `app.js` | Verified correct |
| B8 — `SetWindowStarted/Ended` + `signalLifecycleAnchor` at body_start/body_end/stable | Verified correct |
| N1 — `terminate_when` creates fresh thread per invocation | Verified |
| N4 — `match.event()` and `argsToMatcher` reject non-string kwarg values | Verified |
| N5 — `events.sum()` removed from `AttrNames()` and implementation | Verified |
| N6 — `test()` duplicate name returns error | Verified |
| N7 — `trace.events_within(of=None)` returns all matching events | Verified |
| N8/N9 — exit code comments and RFC §8.6 annotation corrected | Verified |
| N10 — sandbox docs list all denied builtins | Verified |

**LGTM. Clean up the three dead variables post-merge and this is ready.**

PR #118, opened 2026-05-16
feat: RFC-043 Non-deterministic Operators (rc1, v0.13.0)

avatar29A:
## Summary

Ships rc1 of **RFC-043 — Non-deterministic Spec Operators**. Four
small Starlark primitives that give Faultbox specs the vocabulary
P-lang and TLA+ have used for decades to describe a *space of
behaviors*: \`choose()\`, \`nondet()\`, \`halt()\`, \`assume()\`.

rc1 is the **language surface + single-leaf runtime**: every operator
parses, validates, and runs against the current (first-option /
first-leaf) snapshot. The plan-tree fan-out — actually producing
N test runs per \`choose([opts])\` — lands with rc2 alongside
RFC-042 §8.8's body re-execution machinery.

## Commits

| # | Commit | Scope |
|---|---|---|
| 1 | \`2df88c1\` | ChoiceVal + \`choose()\` + \`nondet()\` boolean (§5.1, §5.2) |
| 2 | \`358b113\` | \`halt()\` builtin + \`"halted"\` outcome through SuiteResult/manifest/report (§5.3) |
| 3 | \`aa0569f\` | \`assume(predicate)\` + \`test(assume=[...])\` + sandboxed evaluation thread (§5.4) |
| 4 | \`0b3c05b\` | \`halt()\` outside test body rejected (§5.8 #3) |
| 5 | \`83822b9\` | \`docs/nondeterministic-operators.md\`, spec-language, manifest, RFC flip (§8.8) |

5 PRs, 12 files (excluding two stray docs picked up by \`git add docs/\` —
benign, unchanged from prior epics).

### \`choose([opts])\` / \`choose("name", [opts])\` (§5.2)
Non-deterministic finite choice. rc1 returns the first option; the
\`ChoiceVal\` is recorded on \`rt.choices\` so the plan-tree fan-out
in rc2 can read the option set without re-parsing. Empty lists, non-
list args, non-string names, and wrong arities all error at spec load.

### \`nondet()\` (§5.1)
Sugar for \`choose([True, False])\`. **Overloaded** against the
pre-existing \`nondet(svc, ...)\` variant by argument count — the
service-exclusion form keeps working unchanged. RFC-044 may unify
later.

### \`halt(reason="")\` (§5.3)
Prune the current plan-tree branch. \`RunTest\` catches the \`ErrHalt\`
sentinel before the generic fail path and returns
\`Result="halted"\`. New \`SuiteResult.Halted\` counter +
\`bundle.Summary.Halted\` + HTML report \`"halted"\` outcome (grey pill,
distinct from pass/fail/inconclusive).

\`halt()\` at module top-level or inside \`setup=\` errors at spec
load — guarded via \`rt.inTest\` (atomic.Bool).

### \`assume(predicate)\` + \`test(assume=[...])\` (§5.4)
Plan-tree filter. Top-level form evaluates at spec load (violation
errors the load); per-test form evaluates at body entry (violation
halts the test). Predicates receive a \`choices\` dict mapping each
named \`choose("name", opts)\` call to its currently-selected option.
Predicates run in a sandbox \`starlark.Thread\` tagged \`"assume:<name>"\`;
Starlark errors inside the predicate fail the test instead of halting.

Bare booleans are accepted as a convenience (wrapped in a constant
closure).

## Deferred to rc2 (per the split agreed upstream)

- **§8.5 plan-tree fan-out** — body re-execution / per-leaf
  selection vectors. Lands with RFC-042 §8.8.
- **§8.6 cost-guard + choose-options static check** — no-op in rc1
  (single-leaf means cartesian = 1; Starlark's eager evaluation
  already rejects runtime-computed elements).
- **§8.7 assume predicate AST sandbox** — predicates run in a tagged
  sandbox thread today; the spec-load AST walk that rejects service
  / fault references lands with rc2.
- **§8.9 testops goldens** — unit + CLI tests in rc1; end-to-end
  goldens share the same corpus as the interleaving work.

## Compatibility

- **Specs**: \`fault_matrix\`, \`fault_scenario\`, \`scenario\`,
  \`test()\`, \`def test_*()\` all behave exactly as before. New
  operators are additive; existing specs that never call them see no
  change.
- **\`nondet(svc, ...)\`**: pre-RFC-043 variadic form preserved.
  Zero-arg \`nondet()\` is the new RFC-043 boolean.
- **Bundles**: \`summary.halted\` is omitempty — pre-RFC-043 bundle
  readers see the same fields as before. \`Result="halted"\` is a
  new value alongside \`pass\` / \`fail\` / \`error\` / \`inconclusive\`.

## Test plan

- [x] \`go test ./...\` — green
- [x] \`go vet ./...\` — clean
- [x] choose / nondet arity + type + recording: \`internal/star/choose_test.go\`
- [x] halt sentinel + reason + RunTest path + top-level rejection +
  setup rejection: \`internal/star/halt_test.go\`
- [x] assume top-level + per-test + predicate raise + unnamed-choice
  skip: \`internal/star/assume_test.go\`
- [x] Existing temporal / determinism / plan tests still green
- [x] Manual smoke against a small choose+halt+assume spec

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,8 +50,22 @@ and probability fan-out (§8.9) are deferred to rc2; their flags
 reserved in rc1 with clear "future release" errors so CI integrations
 stay stable. User guide: [docs/exploration.md](docs/exploration.md).
 
-RFC-043 (non-deterministic operators) and RFC-044 (spec-language
-simplification) remain draft RFCs tracked in GitHub issues #112–#113.
+**RFC-043 — non-deterministic operators (rc1).** Four small
+Starlark primitives shipped: `choose([opts])` / `choose("name", [opts])`
+for finite N-way choice, `nondet()` for the non-deterministic boolean
+(sugar for `choose([True, False])`; the pre-existing `nondet(svc)`
+variant for interleaving-control exemption keeps working unchanged),
+`halt(reason="")` for plan-tree branch pruning with a new `"halted"`
+outcome flowing through `SuiteResult`, bundle manifest, and the HTML
+report, and `assume(predicate)` / `test(assume=[...])` for plan-tree
+filtering. rc1 ships the language surface with single-leaf runtime
+semantics — each operator returns the first option / first leaf;
+full plan-tree fan-out and the AST sandbox for assume predicates
+land with rc2 alongside RFC-042 §8.8. User guide:
+[docs/nondeterministic-operators.md](docs/nondeterministic-operators.md).
+
+RFC-044 (spec-language simplification) remains a draft RFC tracked
+in GitHub issue #113.
 RFC-046 carries the post-L1 roadmap (gVisor Path B/C, L4 hermetic
 mode, L5 instruction-boundary research).
 
--- a/README.md
+++ b/README.md
@@ -162,6 +162,7 @@ make demo
 | [Determinism Levels](docs/determinism.md) | L0–L5 taxonomy, the L1 contract, escape hatches (RFC-040) |
 | [Temporal Properties](docs/temporal.md) | `eventually` / `always` / `monitor` / `await_*` / `test()` (RFC-041) |
 | [Plan & Coverage](docs/exploration.md) | `faultbox plan`, `plan.json`, coverage, suggestions (RFC-042) |
+| [Non-deterministic Operators](docs/nondeterministic-operators.md) | `choose()` / `nondet()` / `halt()` / `assume()` (RFC-043) |
 
 ### Tutorial Structure
 
--- a/cmd/faultbox/main.go
+++ b/cmd/faultbox/main.go
@@ -2263,6 +2263,12 @@ func testRowsFromResult(result *star.SuiteResult) []bundle.TestRow {
 			// built on the bundle manifest don't misread a stall as
 			// success.
 			outcome = "inconclusive"
+		case "halted":
+			// RFC-043 §5.3 — body called halt(); plan-tree pruning
+			// signal, not a real verdict. Distinct outcome value so
+			// the report can grey-pill the row without inflating
+			// pass/fail counters.
+			outcome = "halted"
 		default:
 			if tr.FaultBypassed {
 				outcome = "fault_bypassed"
--- a/cmd/faultbox/main.go
+++ b/cmd/faultbox/main.go
@@ -502,13 +502,17 @@ func testStarCmd(starFile string, rcfg star.RunConfig, outputPath, shivizPath, n
 		}
 	}
 
-	// Print summary. Inconclusive count is appended only when >0 so
-	// the line stays unchanged for pre-RFC-041 specs.
+	// Print summary. Inconclusive (RFC-041) and Halted (RFC-043) counts
+	// are appended only when >0 so the line stays unchanged for specs
+	// that don't use either feature.
+	summary := fmt.Sprintf("\n%d passed, %d failed", result.Pass, result.Fail)
 	if result.Inconclusive > 0 {
-		fmt.Fprintf(os.Stderr, "\n%d passed, %d failed, %d inconclusive\n", result.Pass, result.Fail, result.Inconclusive)
-	} else {
-		fmt.Fprintf(os.Stderr, "\n%d passed, %d failed\n", result.Pass, result.Fail)
+		summary += fmt.Sprintf(", %d inconclusive", result.Inconclusive)
 	}
+	if result.Halted > 0 {
+		summary += fmt.Sprintf(", %d halted", result.Halted)
+	}
+	fmt.Fprintln(os.Stderr, summary)
 
 	// Terminal replay hint per failed test (RFC-025 §Observability).
 	// Makes `replay_command` visible without digging into JSON. Skipped
--- a/docs/design/2026-04-22-customer-feedback-analysis.md
+++ b/docs/design/2026-04-22-customer-feedback-analysis.md
@@ -1,0 +1,498 @@
+# inDrive / truck-api PoC — Feedback Analysis & 0.9.x / 0.11.0 Roadmap
+
+**Authored:** 2026-04-22
+**Inputs:** [FAULTBOX_FEEDBACK.md](FAULTBOX_FEEDBACK.md), [FAULTBOX_REPRODUCIBILITY.md](FAULTBOX_REPRODUCIBILITY.md), [TEST_COVERAGE.md](TEST_COVERAGE.md)
+**Validated against:** v0.9.6 (current main)
+**Scope per owner:** finish the asks in the feedback docs in the 0.9.x
+train; cap with **v0.11.0 Reports** (rich HTML artifact per run). **No
+big features** outside the reports or the Reports story.
+
+---
+
+## 0. Executive summary
+
+The customer found 16 real bugs and ranks the core product unique ("the
+seccomp user-notify approach is the reason the PoC was worth doing") but
+says it isn't shippable yet because of three systemic gaps — **gRPC
+proxy on the data path, Linux CI, and docs.** Two of those are
+already substantially closed on current main:
+
+| Customer ask | Status on main | Delta |
+|---|---|---|
+| Native gRPC mocks | **Shipped** (RFC-023, v0.9.0) | Tutorial Ch 18 exists |
+| Proxy on SUT data path | **Shipped** (RFC-024, v0.9.5 + v0.9.6) | Container mode + TCP + benchmarks done |
+| Per-gRPC-method fault targeting | **Shipped** via `error(method="/pkg/Method")` at proxy layer | Needs better surfacing in docs |
+| `fault_assumption` dispatch + `ops=` filter + `fault_zero_traffic` | **Shipped** (v0.9.4) | Customer knows |
+
+The **remaining** asks cluster into five groups:
+
+1. **Documentation overhaul** (customer lost ~47h / 20% of engagement to docs gaps).
+2. **Small missing primitives** — `load_file()`, JWT/JWKS, richer expectations, gRPC error shorthands.
+3. **Reproducibility-by-default** — artifacts dir, replay_command surfacing, version/digest pinning.
+4. **CI / deployment ergonomics** — Linux CI recipe, service-binary rebuild, healthcheck types.
+5. **Processes / commercial** — API stability promise, roadmap visibility, reference customer.
+
+Effort/Profit summary below. **Recommended release sequence:**
+
+- **v0.9.7** — `.fb` archive bundle format (every run → one self-contained file), replay on failure, seed default. Load-time validation.
+- **v0.9.8** — Small primitives: `load_file()`, JWT/JWKS, `expect_*`.
+- **v0.9.9** — Documentation overhaul + CI-on-Linux recipe + Starlark reference.
+- **v0.10.0** — Determinism guardrails (digest pinning, version mismatch detection, `faultbox replay <bundle>`).
+- **v0.11.0** — **Interactive HTML Reports** consumed from a `.fb` bundle (shareable static artifact, SaaS-neutral).
+
+Each release builds on the previous; v0.11.0 is the capstone that
+turns the reproducibility and determinism work into something users
+actually look at.
+
+---
+
+## 1. Validation — what's already shipped vs the reports
+
+The customer reports were written on v0.9.4. Current main is v0.9.6. This
+table catches items the customer may not know are already closed.
+
+| # | Customer ask | Report §ref | Status on v0.9.6 | Evidence |
+|---|---|---|---|---|
+| 1 | Native gRPC mock | FB §3 row 1 | ✅ Shipped v0.9.0 | RFC-023, `@faultbox/mocks/grpc.star`, tutorial Ch 18 |
+| 2 | Proxy on SUT data path | FB §3 row 2 | ✅ Shipped v0.9.5+6 | RFC-024, `preStartProxies`, `host.docker.internal` |
+| 3 | Per-gRPC-method targeting | FB §3 row 3 | ✅ Shipped (surfacing gap) | `error(method="/pkg/Method")` at proxy layer |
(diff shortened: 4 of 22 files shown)


avatar29A (commented):
## Principal Engineer Review — RFC-043 rc1 (Non-deterministic Operators)

The four primitives are well-scoped for an rc1: `ChoiceVal` recording, the `ErrHalt` sentinel, and the `assume` sandbox thread are the right primitives to lock in before rc2 wires fan-out. The overloading of `nondet()` by arg count is pragmatic and backwards-compatible. Two correctness bugs and two doc accuracy gaps need fixing before merge.

---

### 1. `halt()` masks an `always()` invariant violation — priority inversion

**`internal/star/runtime.go` — `RunTest`**

When `alwaysFired` fires first (`cause = TerminationImmediateFail`) and the body simultaneously calls `halt()`, `bo.err` is `ErrHalt`. The guard that catches the halt sentinel is:

```go
if err != nil && cause != TerminationTimeout && cause != TerminationTerminateWhen {
    if errors.Is(err, ErrHalt) {
        return TestResult{Result: "halted", ...}
    }
}
```

`TerminationImmediateFail` is not excluded. The halted early-return fires, the function exits before `monitorErrors` and `Finalize`, and the `always()` invariant violation is silently discarded. The test is recorded as `"halted"` — a no-verdict — instead of `"fail"`.

**Fix:** Restrict the `ErrHalt` early-return to the natural-termination path:

```go
if errors.Is(err, ErrHalt) && cause == TerminationNatural {
    return TestResult{Result: "halted", ...}
}
```

A test that `halt()`s AND has violated an invariant should be `"fail"`, with the halt reason surfaced in the reason field (or dropped). No existing test covers this interaction — add one.

---

### 2. `nondet(svc=x)` bypasses the service-exclusion dispatch via kwarg

**`internal/star/builtins.go` — `builtinNondet`**

The zero-arg / service-arg dispatch is `len(args) == 0`. Keyword arguments are never checked. A caller writing `nondet(svc=my_svc)` has `len(args) == 0`, silently takes the boolean path, returns `True`, and does **not** mark the service as non-deterministic. The service flows back through the interleaving engine as if it were deterministic.

**Fix:** Add before the `len(args) == 0` check:

```go
if len(kwargs) > 0 {
    return nil, fmt.Errorf("nondet(): keyword arguments are not accepted; use nondet(svc) for service exclusion or nondet() for boolean")
}
```

Add a test: `nondet(svc=my_svc)` should error, not return `True`.

---

### 3. `nondet()` GoDoc says "returns False" — contradicts implementation, tests, and user docs

**`internal/star/builtins.go` — `builtinNondet` comment**

The GoDoc line reads:

> "returns False at runtime in v0.13.0-rc1"

The code calls `c.FirstOption()` which is `Options[0]` = `starlark.True`. `docs/nondeterministic-operators.md`, `docs/spec-language.md`, and `TestChoose_ReturnsFirstOption` all say `True`. Change to "returns `True`."

---

### 4. `halt()` inside `setup=` documented as a spec-load error — it is a runtime error

**`docs/nondeterministic-operators.md`**

The doc states:

> "Calls at module top-level or inside `setup=` error at spec load with *'halt() may only be called inside a test body.'*"

`TestHalt_RejectedInSetup` demonstrates this is false for `setup=`: `LoadString` succeeds; the error fires at `RunTest` when the setup callable runs and `inTest.Load()` is still false. This is a runtime `"fail"`, not a spec-load error. The distinction matters for any CI step that runs `faultbox plan` or spec-load validation separately.

**Fix:** "Calls at module top-level error at spec load. Calls inside `setup=` are rejected at run time (before the body starts) with the same message."

---

**N1. `halt()` inside a monitor callback is accepted but surfaces as a confusing "monitor violation"**

`builtinHalt` checks `!rt.inTest.Load()` — false during a test, so `halt()` inside a monitor `check=` or `update=` lambda is allowed. The `*HaltError` propagates to `recordMonitorError`, which stores it as a monitor violation. The user sees `Result="fail"` with reason `"monitor violation: halt"` rather than a clear "halt() cannot be called inside a monitor." Add either a thread-name guard (`strings.HasPrefix(thread.Name, "monitor:")`) or document explicitly that `halt()` in monitor lambdas is unsupported and produces an opaque error.

**N2. `assume` sandbox is name-tagging only — denylist not applied — docs overclaim**

`newSandboxThread("assume:<name>")` tags the thread name but applies no builtin restrictions. An assume predicate can call `fault()`, `service()`, `halt()`, or any runtime builtin. The docs say "runs in a sandboxed evaluation thread" — implies restriction. The PR correctly defers the AST denylist to rc2, but the current text overstates what rc1 enforces. Change to: "rc1: predicates run in a tagged thread; the builtin denylist (applied to monitor predicates) is not yet enforced — lands in rc2."

**N3. Body-time `choose()` invisible to `assume=` predicates — undocumented sharp edge**

`assume=` predicates evaluate before `rt.inTest.Store(true)` and before the body goroutine launches. Any `choose("k", opts)` called inside the body is recorded in `rt.choices` after the predicate has already run. A user writing `assume=[lambda c: c["k"] == 1]` for a choice that only fires inside the body gets a KeyError → `Result="fail"`. Add a note: "Only top-level `choose()` calls (at spec-load time) appear in the `choices` dict passed to `assume=` predicates. Body-level calls are not visible in rc1."

**N4. `assume=` early-exit `TestResult` missing `Matrix` / `ExpectationName` fields**

When `assume=` predicates halt or fail a test at body-entry, the returned `TestResult` omits `Matrix` and `ExpectationName`. These fields are populated after body-entry in the normal path. Harmless in rc1 (single-leaf, matrix info not yet bound), but will silently omit matrix attribution in rc2 when tests belong to a `fault_matrix`. Add a comment flagging the gap; fix in the same rc2 commit that wires matrix cell binding.

**N5. `Halted` count missing from CLI terminal summary**

`cmd/faultbox/main.go` prints `Inconclusive` conditionally in the summary line but never `Halted`. In rc2 with fan-out, a large matrix with many pruned branches will show "0 passed, 0 failed" with no terminal signal that N tests were halted. Add `Halted` alongside `Inconclusive` in the summary output.

**N6. `reportSubtitle` says "All N checks held" when N includes halted tests**

`internal/report/report.go` — `reportSubtitle`: the "All N checks held under the injected faults" path checks only `Failed == 0 && Errored == 0`. If N tests are halted, `Total` includes them and the subtitle is misleading — halted tests produced no verdict. Add a branch for `Halted > 0` or exclude halted from the "all held" count.

**N7. `spec-language.md` outcome table does not list `"halted"`**

The spec-language reference table of `TestResult.Result` values (`"pass"`, `"fail"`, `"error"`, `"inconclusive"`, ...) does not include `"halted"`. This is the canonical reference consumed by LLM agents and CI tooling parsing bundle output. Add a row.

---

## QUESTIONS

**Q1. `ChoiceVal.Truth()` — is `ChoiceVal` ever a user-visible Starlark value?**

`builtinChoose` and `builtinNondet` both return `FirstOption()` (a raw `starlark.Bool`), not the `ChoiceVal` wrapper. `ChoiceVal.Truth()` is therefore unreachable in rc1. In rc2, if fan-out returns `ChoiceVal` to Starlark callers, `Truth()` returns truthy for any non-empty choice — making `if choose([False, True])` always-truthy regardless of the first option. Clarify the rc2 API contract for `ChoiceVal` as a user-visible value, and add a comment on `Truth()` documenting the intended semantics.

**Q2. `halted` filtered by `FailOnly`; `inconclusive` is not — intentional?**

In `RunAll`, `inconclusive` results unconditionally append to `suite.Tests` even with `cfg.FailOnly=true`. `halted` results are filtered by `!cfg.FailOnly`. The asymmetry means `--show=fail` gives users inconclusive tests (likely wanted) but no halted tests (likely also wanted for debugging pruned branches). Intentional?

**Q3. `assume` predicate error → `Result="fail"` rather than `Result="error"`**

A `KeyError` or `AttributeError` inside an `assume=` predicate maps to `"fail"`. This lumps a spec-authoring bug (wrong `choices` key) into the same bucket as a genuine SUT behavioral failure. Should predicate errors map to `"error"` (the panic / unexpected-exception bucket) to distinguish them from real assertion failures?

---

## Summary

| Severity | Count |
|----------|-------|
| Blocking | 4 |
| Non-blocking | 7 |
| Questions | 3 |

B1 (halt masks always-violation) is the most consequential — it silently discards an invariant failure as a no-verdict result. B2 (kwarg bypass) and B3 (wrong GoDoc) are one-liners. B4 (setup= doc) is one sentence. All four are surgical fixes requiring no design changes.

avatar29A:
## Review responses — fixes in b617b45

All 4 blocking and 7 non-blocking items addressed in one commit. Summary below; details in the commit message.

### Blocking

- **B1 — halt() priority inversion:** two-line fix in `runtime.go`. The cause-escalation block (line 1235) that promoted any body error to `TerminationImmediateFail` is now skipped for `ErrHalt` (so a clean halt() stays `TerminationNatural`), and the halted early-return is gated on `cause == TerminationNatural`. The combination means: a body that only halts → `halted`; a body that halts AND violated `always()` → `fail` (the always-violation cause already set `TerminationImmediateFail` via the select arm before the halt is observed).
- **B2 — `nondet(svc=...)` kwarg bypass:** rejected explicitly; new test `TestNondet_RejectsKwargs` verifies the error message and that the service stays untagged.
- **B3 — GoDoc:** fixed to "returns True".
- **B4 — setup= halt() doc:** rewritten to spell out the runtime-vs-load distinction.

### Non-blocking

- **N1 — halt() in monitor:** documented as unsupported (opaque "monitor violation: halt"); recommend `assume()` to filter branches. Code-level guard deferred to rc2 where monitor sandboxing gets its full AST walk.
- **N2 — assume sandbox overclaim:** doc tightened to "tags the thread name; denylist lands in rc2."
- **N3 — body-time choose() invisible to assume=:** documented the sharp edge.
- **N4 — Matrix/ExpectationName on assume early-exit:** `runtime.go` now snapshots `currentFaultScenario.Matrix` and `expectationName` before the assume loop and threads both into the early-return TestResult.
- **N5 — Halted in CLI summary:** appended alongside `Inconclusive` when > 0.
- **N6 — reportSubtitle:** `verdictTotal := Total - Halted` excludes halted leaves from "all held" framing; a halt-only run reads "No verdicts — all N branches halted." instead.
- **N7 — outcome table:** added `halted` and `inconclusive` rows.

### B1 e2e test

Did not add an end-to-end test for the halt+always-violation interaction. Reproducing it deterministically requires (a) `always(lambda t: False)` registered before body entry, (b) an event arriving before halt() to trip `alwaysFired`, and (c) the body then halting. There's no Starlark builtin to emit synthetic events from inside a test body, and timing the violation against `halt()` from outside is racy. Documented the priority semantics inline in the code comment; the proper coverage lands with rc2's testops goldens for plan-tree pruning.

### Question answers

**Q1 — `ChoiceVal.Truth()` reachability and rc2 contract.** Correct that rc1 never exposes `ChoiceVal` to Starlark callers — both builtins call `FirstOption()` and return the unwrapped value. For rc2, the design intent is that fan-out happens at the *plan engine* level, not by returning a `ChoiceVal` to the body. Each leaf gets one re-execution with a concrete option. `ChoiceVal` stays internal to the plan-tree representation. The `Truth()` method exists only because Starlark's `Value` interface demands it; I'll add a `// rc2: ChoiceVal must not leak to user code` comment next slice.

**Q2 — halted vs inconclusive `FailOnly` asymmetry.** Not intentional — historical artifact of `inconclusive` predating RFC-043. The `--show=fail` reader probably wants both as "interesting non-pass results." I'll align `halted` to also bypass `FailOnly` in the rc2 cleanup commit; treating it as a behavioral fix rather than rolling it into this review pass to keep the diff focused on the four blocking items.

**Q3 — predicate error → `"error"` vs `"fail"`.** Good call. A spec-authoring bug (wrong `choices` key) is morphologically the same as a panic, not an assertion failure. Will change to `"error"` in the same rc2 commit that adds the AST denylist (§8.7) — both touch the assume evaluation path.

Both Q2 and Q3 fixes are small but the rc2 plan-tree work will rewrite this code substantially; bundling them avoids merge churn.

Test status: `go test ./... && go vet ./...` clean on b617b45.

avatar29A (commented):
## Follow-up Review — commit \`b617b453\`

Verified all 4 blocking and 7 non-blocking fixes. All are correctly implemented.

### What verified

| Issue | Status |
|-------|--------|
| B1 — `halt()` priority: cause-escalation excludes `ErrHalt`; early-return gated on `cause == TerminationNatural` | Verified correct |
| B2 — `nondet(svc=x)` kwarg guard + error message + test asserting service is NOT marked | Verified correct |
| B3 — GoDoc "returns False" → "returns True" | Verified correct |
| B4 — `halt()` in `setup=` documented as runtime rejection, not spec-load error; monitor-lambda behaviour noted | Verified correct |
| N1 — `halt()` in monitor callback documented with opaque-error warning + `assume()` redirect | Verified correct |
| N2 — `assume` sandbox doc: "no denylist enforced in rc1" explicit; rc2 AST walk cross-referenced | Verified correct |
| N3 — Body-time `choose()` invisible to `assume=` documented with rc1 visibility callout block | Verified correct |
| N4 — `assume=` early-exit `TestResult` captures `preMatrix` / `preExpect` before the loop | Verified correct |
| N5 — CLI summary appends `", N halted"` via the new string-builder path | Verified correct |
| N6 — `reportSubtitle` uses `verdictTotal = Total - Halted`; new "No verdicts — all N branches halted" case | Verified correct |
| N7 — `spec-language.md` outcome table now lists `"halted"` (RFC-043) and `"inconclusive"` (RFC-041) | Verified correct |

### One narrow race to track (pre-existing, not introduced by this fix)

The `case bo = <-bodyDone:` arm in the RunTest select has no secondary drain for `alwaysFired`. If both channels are ready simultaneously and Go's scheduler picks `bodyDone`, an `always()` violation is missed and the halted-body path fires. This race existed before RFC-043 and was not introduced or worsened by the B1 fix — B1 correctly handles the primary case where `alwaysFired` wins the select. Adding a secondary drain to the `bodyDone` arm (mirroring what the `bodyCtx.Done()` arm already does) would close the window entirely; worth a follow-up issue rather than a blocker.

**LGTM. Ready to merge.**

PR #120, opened 2026-05-16
fix(runtime): drain alwaysFired/twFired on bodyDone arm (#119)

### LINKED ISSUE #119: bug: RunTest select races bodyDone vs alwaysFired — always() violation can be missed
## Issue

In `internal/star/runtime.go` the `RunTest` select arms `bodyDone`, `alwaysFired`, `twFired`, and `bodyCtx.Done()`. If both `bodyDone` and `alwaysFired` are ready at the same scheduling tick and the Go runtime picks `bodyDone`, the `always()` invariant violation is dropped — there is no secondary drain on that arm (unlike the `bodyCtx.Done()` arm, which does drain `alwaysFired` / `twFired`).

Flagged by principal-engineer review on PR #118 (RFC-043 rc1): pre-existing, not introduced or worsened by the halt() priority-inversion fix landed in b617b45.

## Proposed fix

After `bo = <-bodyDone`, drain `alwaysFired` and `twFired` non-blockingly and update `cause` accordingly — mirror the disambiguation already present in the `bodyCtx.Done()` arm.

```go
case bo = <-bodyDone:
    // Body completed under its own power. Drain always/tw so a
    // simultaneous violation isn't dropped by scheduler choice.
    select {
    case <-alwaysFired:
        cause = TerminationImmediateFail
    case <-twFired:
        cause = TerminationTerminateWhen
    default:
    }
```

## Impact

Low: requires a body that emits an `always()`-violating event in the same scheduling tick as it returns. Not reproducible in our existing test suite (which is why this slipped through pre-RFC-043). A determinism-stress test that emits N events and returns immediately would expose it.

## Related

- PR #118 follow-up review
- RFC-041 §5.2 (always() semantics)
- RFC-043 §5.3 (halt() interaction with always() — depends on this fix being correct)

avatar29A:
## Summary

- Closes #119 — RunTest's `select` had no secondary drain on the `bodyDone` arm. When both `bodyDone` and `alwaysFired` (or `twFired`) were ready in the same scheduling tick, Go's select chose uniformly at random, silently dropping invariant violations / terminate_when signals when `bodyDone` won.
- Fix: non-blocking drain of `alwaysFired` and `twFired` right after `bo = <-bodyDone`. Mirrors the disambiguation already present in the `bodyCtx.Done()` arm.

## Why no test

Reproducing a Go-select race deterministically requires either (a) stress-looping 100+ times and asserting zero flakiness, which is slow and probabilistic, or (b) injecting test-only channel pre-arming, which would alter the code under test. The fix is structurally identical to the `bodyCtx.Done()` drain that's been correctness-checked since RFC-041 landed — copying its pattern is the proof.

## Test plan

- [x] `go test ./... && go vet ./...` clean
- [x] No behavioral change to any of the existing always()/terminate_when test paths (drain only fires when signal is already ready, which the existing tests don't construct)

### DIFF
--- a/internal/star/runtime.go
+++ b/internal/star/runtime.go
@@ -1199,6 +1199,20 @@ func (rt *Runtime) RunTest(ctx context.Context, name string) TestResult {
 		// Body completed under its own power. cause stays Natural
 		// unless the body raised; that's classified after we examine
 		// the error below.
+		//
+		// Drain alwaysFired / twFired non-blockingly to disambiguate
+		// the simultaneous-ready case: if both bodyDone and one of the
+		// signals were ready at the same scheduling tick, Go's select
+		// chooses uniformly at random — losing the violation/terminate
+		// signal would silently drop an invariant failure (issue #119).
+		// Mirrors the disambiguation in the bodyCtx.Done() arm.
+		select {
+		case <-alwaysFired:
+			cause = TerminationImmediateFail
+		case <-twFired:
+			cause = TerminationTerminateWhen
+		default:
+		}
 	case <-alwaysFired:
 		// always() invariant violated mid-run. RFC-041 §5.2 — fail
 		// immediately, rolling pending eventually's into the verdict.


avatar29A (commented):
LGTM. The drain is correct, minimal, and consistent with the `bodyCtx.Done()` arm. Priority order (`alwaysFired` before `twFired`) matches the invariant-beats-terminate_when precedence established in RFC-041. Interacts cleanly with the B1 halt-masking fix from #118 — draining `alwaysFired` sets `cause = TerminationImmediateFail`, which blocks the `ErrHalt && cause == TerminationNatural` guard and prevents a concurrent `halt()` from silencing the violation.

PR #121, opened 2026-05-16
feat: RFC-042 rc2 partial — body re-execution engine + probability fan-out

avatar29A:
## Summary

Ships the v0.13.0-rc2 body-re-execution engine that turns RFC-042's plan tree and RFC-043's `choose()` axes from analysis-only artifacts into actual multi-leaf test execution.

- **Plan-tree fan-out for named `choose()` axes (RFC-043 §5.2).** `choose("retries", [0, 1, 3])` now produces 3 separate test executions; each leaf carries a stable `LeafID` through `TestResult` → bundle manifest → HTML report.
- **§8.9 syscall-level probability fan-out.** `delay()` / `deny()` accept `max_fires=N` and `mode="exhaustive"|"stochastic"`. Exhaustive mode (the new default) fans out 2^N leaves per rule with a deterministic per-occurrence fired/not-fired vector replacing the seeded-RNG decision. Stochastic mode preserves the legacy behavior verbatim.
- **Mixed-radix cross-product.** Choice axes × probability sites combine into one enumeration; a spec with `choose("k", [a,b])` and a `probability=0.3, max_fires=2` rule produces 2 × 4 = 8 leaves.
- **Multi-leaf bundle attribution.** `TestResult.LeafID` flows into `bundle.TestRow.LeafID` and the HTML report's tests table suffixes the display name with `"[leaf N]"`. Single-leaf executions stay byte-identical to rc1.

## What lands per PR

| PR | Scope | LOC |
|----|---|---|
| 1/5 | `PlanLeaf` type, `RunTestLeaf` entry point, `ChoiceVal.Selected(leaf)`, `currentLeaf` field | ~180 |
| 2/5 | Plan-tree enumerator (`enumerateLeaves`), body re-exec loop (`runTestFanout`), `RunAll` integration, spec/body choice partitioning | ~440 |
| 3/5 | `max_fires=`/`mode=` kwargs, `PlanLeaf.ProbabilityOutcomes`, `engine.FaultRule.Mode/MaxFires`, `SessionConfig.ProbabilityDecider`, `launch_linux.go` consultation, site recording in `applyFaults` | ~555 |
| 4/5 | `bundle.TestRow.LeafID`, CLI/report wiring | ~50 |
| 5/5 | RFC-042 status flip, spec-language.md kwarg tables, nondeterministic-operators.md rewrite, feature-manifest rows, CHANGELOG | ~50 |

## Deferred (separate slices)

- **§8.8 spec-level interleaving execution** — A2. Heaviest engine work remaining in v0.13.0; the body-re-execution substrate lands here for it to plug into.
- **Protocol-level probability fan-out** — `response()` / `error()` / `drop()` need the same surface threaded through `ProxyFaultDef` and `internal/proxy`.
- **Static trigger-count analysis** — today exhaustive mode requires `max_fires=N`; unmodeled rules fall back to stochastic without a plan warning.
- **`unmodeled_fanout` plan diagnostic** — surfaced in `faultbox plan` once static analysis lands.
- **Per-leaf drill-down navigation / `--leaf <id>` replay** — A3.

## Test plan

- [x] `go test ./... && go vet ./...` clean on every PR commit (CI green throughout)
- [x] 14 new unit tests covering: leaf-pinned choose() selection, anonymous-choose() nil-leaf fallback, cross-product enumeration, mixed-radix probability encoding (2^max_fires), site dedup by key, runtime decider closure end-to-end, kwarg parsing + validation (7 cases), multi-leaf bundle row shape
- [x] No behavior change for specs that don't use named `choose()` or `max_fires=` (regression-tested by every pre-existing test passing unchanged — single-leaf path stays nil-leaf throughout)

## Migration

Existing specs with `probability=p` keep working unchanged. Adding `max_fires=N` opts into 2^N exhaustive fan-out; adding `mode="stochastic"` keeps the legacy single-realization RNG path. The default for new `max_fires=` declarations is `"exhaustive"`. RFC §8.11 migration notes documented in [docs/exploration.md](docs/exploration.md).


avatar29A (commented):
## Principal Engineer Review — RFC-042 rc2 partial (body re-execution + probability fan-out)

The structural decisions are right: `PlanLeaf` as the pinning unit, the discovery-run → fanout loop, mixed-radix cross-product, `SessionConfig.ProbabilityDecider` injection. Six issues need fixing before merge — four Go correctness bugs and two doc accuracy gaps that would send spec authors in the wrong direction.

---

### 1. Data race on `rt.currentLeaf`

**`internal/star/runtime.go` / `internal/star/leaf.go` — `RunTestLeaf`, `probabilityDecider`**

`rt.currentLeaf` is a plain `*PlanLeaf` field. `RunTestLeaf` writes it (and the deferred nil-clear) from the RunAll goroutine; `probabilityDecider` reads it from the engine's per-notification handler goroutines (spawned from `launch_linux.go` per seccomp event). For specs with reused services the notification loop stays alive between leaves — the `defer rt.currentLeaf = nil` at the end of leaf N races with the next leaf's assignment and with concurrent decider reads mid-leaf.

**Fix:** Add a `currentLeafMu sync.RWMutex`; writer-lock in `RunTestLeaf`, reader-lock in `probabilityDecider` and anywhere else `currentLeaf` is read.

---

### 2. `NextProbabilityOccurrence` lazy-init data race

**`internal/engine/fault.go` — `NextProbabilityOccurrence`**

```go
func (r *FaultRule) NextProbabilityOccurrence() int {
    if r.probCounter == nil {
        r.probCounter = &atomic.Int64{}   // ← two goroutines can both see nil
    }
    n := r.probCounter.Add(1)
    return int(n) - 1
}
```

Two concurrent notification-handler goroutines on the same `*FaultRule` can both observe `probCounter == nil`, both allocate, and one allocation is lost — the surviving pointer starts at 1, not 0, producing a wrong occurrence index for the exhaustive outcome vector.

**Fix:** Initialize `probCounter` in `SetDynamicFaultRules` where the rule is first established, or use `sync.Once`.

---

### 3. Syscall-family expansion silently multiplies fan-out sites

**`internal/star/runtime.go` — `recordProbFault`**

When a `FaultDef` has no `Label`, its syscall expands to a family (e.g. `"write"` → `write`, `writev`, `pwrite64`). `recordProbFault` records a separate site key per expanded syscall — `"svc:write"`, `"svc:writev"`, `"svc:pwrite64"` — and `enumerateLeaves` treats each as an independent probability axis. A single unlabeled `deny("EIO", probability=0.3, max_fires=2)` on a write-class fault produces 2² × 2² × 2² = 64 leaves instead of 4. The ProbabilityDecider dispatches to the per-expanded-syscall counter, misaligning the occurrence vectors across the three rules. The spec author has no way to know this happened — no warning, no plan diagnostic.

**Fix (minimum):** Use the pre-expansion fault key (`svcName + ":" + unexpandedSyscall`) when no label is set, recording once per `FaultDef` not once per expanded variant. Or reject unlabeled `max_fires=` at spec load with: `"max_fires= on an unlabeled fault that expands to multiple syscalls is ambiguous; add label=\"...\" to scope it"`.

---

### 4. `mode="exhaustive"` without `max_fires=` silently behaves as stochastic

**`internal/star/builtins.go` — `deny()` / `delay()` kwargs**

`deny("EIO", probability=0.5, mode="exhaustive")` is accepted at spec load with `MaxFires=0`. In `recordProbFault`, `site.MaxFires <= 0` causes an early return — no fan-out site is recorded. The engine carries `Mode="exhaustive"` on the rule and a `ProbabilityDecider` is attached, but the decider returns un-pinned on every call (no `ProbabilityOutcomes` entry) and falls through to the RNG. The spec author explicitly requested exhaustive mode and silently gets stochastic behaviour.

**Fix:** Return a spec-load error: `"mode=\"exhaustive\" requires max_fires= > 0"`. The converse case — `mode="stochastic"` without `max_fires=` — is fine (no fan-out was ever intended). The rejection test matrix in `TestProbabilityFanout_RejectsBadKwargs` needs a case for this path.

---

### 5. `docs/exploration.md` not updated — migration language overstates disruption

**`docs/exploration.md` — "What's reserved for rc2" section**

This file was not modified in the PR. The section still reads in future tense:

> "`fault(..., probability=p)` keeps today's stochastic semantics in rc1; the exhaustive fan-out behavior switches on in rc2. When rc2 lands, existing specs see a behavior change."

rc2 has landed. The "behavior change" claim is also **wrong**: specs that use only `probability=p` with no `max_fires=` are completely unaffected — `recordProbFault` early-returns on `MaxFires <= 0` and no fan-out sites are recorded. Exhaustive mode only activates when the author explicitly adds `max_fires=N`. The doc should be updated to:

1. Replace the future-tense section with the actual rc2 story.
2. Correct the migration claim: "only specs that explicitly add `max_fires=N` enter exhaustive fan-out; bare `probability=p` declarations remain stochastic without any edit."

The CHANGELOG has the accurate text; the user guide needs to match it.

---

### 6. Composition example in `nondeterministic-operators.md` claims assume-pruning is live in rc2

**`docs/nondeterministic-operators.md` — composition example**

The updated example summary says "assume-pruned branches show up as 'halted' too" — implying assume-pruning for `choose()` fan-out is implemented in rc2. The "Out of scope" section three lines below correctly defers "Body re-execution / plan-tree fan-out for `assume=`" to a follow-up. The two paragraphs contradict each other.

Additionally, the example predicate `lambda choices: choices["retries"] < 5` references `"retries"` — a choice declared inside `def scenario()` (body-time). Per the documented rc1/rc2 limitation, body-time choices are absent from `choices` at predicate evaluation time. The example would raise `KeyError` → `Result="fail"`, not a clean prune. The example and the surrounding claim must be corrected before spec authors follow it.

---

**N1. `enumerateLeaves` doc comment wrong about bit-0 semantics**

Comment says "bit 0 = first option which is `true`" — bit 0 of `digit=0` is `(0>>0)&1 = 0` (false). Leaf 0 encodes "every occurrence does NOT fire." `TestEnumerateLeaves_ProbabilityAxes` correctly asserts all-false for leaf 0, contradicting the comment. Fix the comment.

**N2. `runTestFanout` stop-on-first-leaf-fail semantics undocumented**

All leaves run unconditionally. If leaf 0 panics or crashes the service, leaf 1 may execute against degraded state. Document the intended behaviour and consider a `RunConfig.StopOnFirstLeafFail` knob for rc2.

**N3. `PlanLeaf.Index` comment describes wrong `LeafID` format**

Comment says stable `LeafID` like `"test_foo[1]"` — actual format is a bare integer string `"0"`, `"1"`, etc. Fix the comment; add a note that it is 0-indexed.

**N4. Seed stamp on exhaustive-mode leaf rows is meaningless**

`leafResults[i].Seed = seed` records the suite-level RNG seed in every leaf result. For exhaustive probability leaves the outcome is pinned — the seed is irrelevant and misleads consumers trying to replay "leaf 3 of seed X." Consider omitting `Seed` from exhaustive-mode rows, or documenting the constraint.

**N5. `probability=` type column says "string" in spec-language.md tables**

New code examples use `probability=0.5` (float). The kwarg table still says type is `"string"`. Update to `"string or float (0.0–1.0 or \"50%\")"` for both `delay()` and `deny()` tables.

**N6. `deny()` migration note misleads spec authors into auditing existing specs**

Spec-language.md `deny()` table: "Required to be `stochastic` for specs that rely on pre-rc2 behavior." As stated in B5, bare `probability=p` is unaffected — authors don't need `mode="stochastic"` unless they also add `max_fires=`. Change to: "Use `mode=\"stochastic\"` to keep RNG semantics when also adding `max_fires=N`."

**N7. No `RunAll`-level test for choice × probability cross-product**

`TestEnumerateLeaves_ChoiceAndProbCrossProduct` exercises the leaf enumeration directly. There is no `RunAll`-level test verifying that the ProbabilityDecider is correctly wired and consulted for the mixed axis case. Add one when the service-start scaffolding permits.

**N8. No test for `max_fires=` on a rule that never fires during the body run**

A declared `max_fires=2` site that the SUT never reaches still produces 4 leaves, all yielding identical results. Document and test this case to prevent future regressions when a "no-op leaf dedup" optimization is added.

---

## QUESTIONS

**Q1. Conditional `choose()` branches — silent axis drop on the discovery run**

`collectNamedAxes` deduplicates on name, taking the first occurrence. If the same name appears in two conditional branches, only the branch taken during the discovery run is recorded. The other branch's options are silently dropped. Is this documented? Should conflicting option-sets be a spec-load error?

**Q2. `mode=""` stored vs `"exhaustive"` documented default**

`deny(..., max_fires=2)` stores `Mode=""` on `FaultDef` / `FaultRule`; the engine treats `Mode != "stochastic"` as exhaustive. The spec-language table says default is `"exhaustive"`. The internal sentinel `""` vs the public value `"exhaustive"` is an implementation detail but worth a comment to prevent a future `Mode == "exhaustive"` equality check from breaking when `""` is stored.

**Q3. Stochastic leaf `probCounter` never advances**

For `mode="stochastic"` rules with a `ProbabilityDecider` attached, the decider is not consulted and `NextProbabilityOccurrence` is also not called — `probCounter` stays at 0 across all leaves of the same test. This is correct (stochastic rules don't use occurrence index), but the comment says "`probCounter` tracks decisions independently." Clarify the comment to say stochastic rules don't update the counter.

---

## Summary

| Severity | Count |
|----------|-------|
| Blocking | 6 |
| Non-blocking | 8 |
| Questions | 3 |

The two races (B1, B2) are detectable with `go test -race` against a reused-service spec with exhaustive probability fan-out and will produce wrong occurrence indices or incorrect outcome vectors. B3 (syscall-family explosion) is a silent semantic bug that will manifest for any spec using an unlabeled fault on a write-class syscall with `max_fires=`. B4 (exhaustive-without-max_fires no-op) misleads authors. B5 and B6 are doc errors that will cause spec authors to do unnecessary migration work and copy broken example code.

avatar29A:
## Review responses — fixes in 9122fea

Plus the CI build error from the macOS/Linux split is now fixed (was passing `&rule` where `rule` was already `*FaultRule`; the Linux-only file isn't compiled on the macOS host so I missed it locally).

### Blocking

- **B1 — currentLeaf data race:** added `currentLeafMu sync.RWMutex` on Runtime. `RunTestLeaf` swaps `currentLeaf` under the writer lock; `probabilityDecider` and `builtinChoose` read via a new `snapshotCurrentLeaf` accessor under the reader lock. `go test -race ./...` now clean across `internal/star` and `internal/engine`.
- **B2 — counter lazy-init race:** pre-initialize `counter` and `probCounter` in `SetDynamicFaultRules` (session.go) and the static-rule setup (launch_linux.go). The lazy paths in `ShouldFire` / `NextProbabilityOccurrence` stay as defensive fallback but no longer race in the notification loop.
- **B3 — syscall-family expansion:** moved fan-out site recording out of the per-expanded-syscall loop AND above the nil-session bail-out, so a single unlabeled `write=deny(probability=, max_fires=)` records exactly one `ProbFaultSite` (keyed on pre-expansion syscall) instead of three. Also override `rule.Label` for all expanded siblings so the decider dispatches to the shared key. New test `TestProbabilityFanout_FamilyExpansionRecordsOneSite` pins this.
- **B4 — mode=exhaustive without max_fires:** rejected at spec load with `mode=\"exhaustive\" requires max_fires= > 0`. New case in `TestProbabilityFanout_RejectsBadKwargs`.
- **B5 — docs/exploration.md:** rewrote the "What's reserved for rc2" section into "What landed in rc2" + "What's still reserved (post-rc2 follow-ups)". Migration note corrected: bare `probability=p` is unaffected; `max_fires=N` is the explicit opt-in to exhaustive fan-out.
- **B6 — nondet doc composition example:** rewrote so `retries` and `fault` are declared at module scope (visible to `assume=`); corrected summary to flag that `assume=`-driven plan-tree pruning is still deferred (matches "Out of scope" section three lines below).

### Non-blocking folded into the same commit

- **N1 (enumerateLeaves comment):** corrected from "bit 0 = first option which is `true`" to "no occurrence fires" — matches `TestEnumerateLeaves_ProbabilityAxes`.
- **N3 (PlanLeaf.Index format):** corrected from `test_foo[1]` to bare integer string, noted 0-indexed.
- **N5 (probability= type):** delay/deny tables now declare type as "string or float".
- **N6 (mode= migration note):** corrected to "bare probability=p is unaffected; you only opt in by adding max_fires=N."

### Non-blocking deferred to follow-ups

- **N2** (stop-on-first-leaf-fail knob), **N4** (omit Seed from exhaustive-mode rows), **N7** (RunAll-level cross-product test — needs service scaffolding), **N8** (zero-fire `max_fires=` dedup) — tracked as separate issues to keep this fix-pass focused on the blockers + the doc/comment items that mislead readers today.

### Question answers

**Q1 — conditional `choose()` branches.** Correct: `collectNamedAxes` dedupes by name, taking the first occurrence. If the discovery run takes branch A and a later leaf takes branch B that calls `choose("k", different_opts)`, the axis recorded reflects only A's option set. **Decision:** document the limitation in the nondet doc's "Out of scope" + emit a spec-load warning (not error) when the AST walk detects multiple `choose("samename", ...)` call sites with different option sets. The error stance is too strict — refactoring an axis into a shared helper is legitimate. Will land in a follow-up before v0.13.0 ships.

**Q2 — Mode="" vs "exhaustive" sentinel.** Good catch. The internal representation stores `""` when the user didn't pass `mode=`, and `"exhaustive"` only when they passed it explicitly. The engine checks `Mode != "stochastic"`, so both `""` and `"exhaustive"` take the exhaustive path. Added a comment on the `Mode` field; an explicit equality check against `"exhaustive"` would silently regress today's specs that omit the kwarg. Worth normalizing to `"exhaustive"` at parse time in a small follow-up so the internal contract matches the documented default.

**Q3 — stochastic leaf probCounter never advances.** Correct — `Mode == "stochastic"` short-circuits the decider consultation and `NextProbabilityOccurrence` is never called. The comment was overgeneral; will tighten it in the same follow-up that normalizes Mode (touches the same code paths).

### Test status

- `go test -race ./...` clean (B1/B2 were racing before this commit)
- `GOOS=linux go build ./...` clean (fixes CI)
- `go vet` clean on both platforms
- 2 new test cases (B3 + B4)

CI should go green on the push.

avatar29A (commented):
## Re-review: commit 9122fead — LGTM

All six blocking issues are correctly addressed.

**B1 — `currentLeaf` data race** ✅  
`currentLeafMu sync.RWMutex` added to `Runtime`; `RunTestLeaf` writer-locks the swap and the nil-reset in the defer; `snapshotCurrentLeaf()` reader-locks the read. Both `builtinChoose` and `probabilityDecider` now go through the accessor. Race window is closed.

**B2 — `probCounter` / `counter` lazy-init race** ✅  
Both fields are pre-initialized in `SetDynamicFaultRules` before the rules slice is handed to the kernel, and again in the `ruleMap` loop. No path can now race on a nil pointer.

**B3 — family-expansion multiplies probability sites** ✅  
Recording loop moved above the nil-session bail-out and iterates the pre-expansion `faults` map (one entry per `FaultDef`). All expanded siblings get the same `rule.Label` so they share a single `ProbFaultSite`. Regression test `TestProbabilityFanout_FamilyExpansionRecordsOneSite` covers exactly this.

**B4 — `mode="exhaustive"` without `max_fires=` silently misroutes** ✅  
Spec-load error added in `builtins.go`; `TestProbabilityFanout_RejectsBadKwargs` extended with the new case.

**B5 — `exploration.md` described planned features as shipped** ✅  
Rewritten to present tense; "What landed in rc2" / "What's still reserved" sections make the boundary clear; migration note corrected.

**B6 — nondeterministic-operators composition example broken** ✅  
Choices moved to module scope, `.get()` predicate used correctly, summary paragraph updated.

---

**One pre-existing non-blocking note** (not introduced by this fix):

The B3 recording loop iterates `map[string]*FaultDef`, which is non-deterministic in Go. If a spec declares two or more probability faults on the same service, the order `recordProbFault` appends sites is random across runs. `enumerateLeaves` sorts by `LeafID` which encodes the site index, so unstable insertion order → unstable `LeafID` → potentially different leaf identity between `faultbox plan` and `faultbox run`. This predates the B3 fix (original code had the same map range). A `sort.Slice` on the recorded sites before `enumerateLeaves` would pin it. Worth a follow-up ticket but not a blocker for this rc.

**LGTM** — rc2 is ready to merge.

PR #123, opened 2026-05-26
feat: RFC-042 §8.8 — parallel(interleavings=) plan-tree fan-out

avatar29A:
## Summary

Ships RFC-042 §8.8 — spec-level interleaving fan-out via `parallel(..., interleavings=)`. Third fan-out axis to land in v0.13.0-rc2 (alongside named `choose()` and syscall-level probability fan-out from PR #121).

- **Kwarg surface.** `parallel(fn1, fn2, ..., interleavings=)` accepts `1` (default, single execution — preserves rc1 semantics), `"all"` (full factorial), `"critical"` (2N-1 heuristic), integer `N` (capped subset). Reserved values `"dpor"` and `"sut-internal"` surface explicit "future release" errors citing RFC-009 / RFC-040 Appendix A.
- **Plan walker integration.** Every `parallel()` call site is recorded during the body's discovery run (keyed on file:line); the enumerator folds the per-policy cardinality into the mixed-radix cross-product alongside choice + probability axes.
- **Engine wiring.** When a leaf pins an `InterleavingIndex` for a site, `parallelWithLeaf` launches the branches in the per-leaf order. Sites without a leaf pin (single-policy or pre-rc2 calls) fall through to the existing `parallelSimple` / `parallelWithExplore` paths.

## Scope limit (rc2)

Today's "interleaving" is **launch ordering** — branches launch sequentially in the per-leaf order. Mediated-event-level interleaving (concurrent branches with the engine releasing their syscalls in a specific sequence via RFC-014's hold queue) is a follow-up. The kwarg surface + leaf descriptors + plan-walker integration are the substrate that work plugs into.

This is documented in the RFC §8.8 status header, the user guide (`docs/exploration.md`), the spec-language reference, and the CHANGELOG.

## What lands per PR

| PR | Scope | LOC |
|----|---|---|
| 1/5 | `interleavings=` kwarg + `InterleavingPolicy` taxonomy + reserved-value rejections | ~240 |
| 2/5 | Body-time `parallel()` site recording (file:line dedup, reset between leaves) | ~220 |
| 3/5 | `PlanLeaf.InterleavingIDs` + enumerator fold-in + RunAll integration | ~300 |
| 4/5 | `permutationByIndex` + `interleavingOrdering` + `parallelWithLeaf` engine path | ~205 |
| 5/5 | RFC §8.8 status flip, exploration.md, spec-language.md, feature manifest, CHANGELOG | ~50 |

## Test plan

- [x] `go test -race ./...` clean (no new races introduced)
- [x] `go vet ./...` clean
- [x] 17 new unit tests covering: kwarg parsing (5 accepted forms + 6 rejections), policy cardinality (single/all/n/critical edge cases), site recording + dedup, reset semantics, Lehmer-code permutation decoding (full enumeration for n=2,3), per-policy ordering mapping, end-to-end RunAll fan-out (2-leaf and single-leaf back-compat), accessor contract
- [x] No behavior change for specs that don't pass `interleavings=` (regression guard: pre-existing `parallel()` tests pass unchanged — default policy collapses to single-leaf)

## Deferred follow-ups (tracked)

- Mediated-event-level interleaving execution (the headline post-rc2 piece)
- `wait_all` / `wait_n` / `wait_first` builtins (likely RFC-044)
- "critical" policy refinement (exact subset definition lands with RFC-009)
- Composition with `nondet(svc)` exemption (currently the two surfaces are independent)


avatar29A (commented):
## Review: RFC-042 §8.8 — `parallel(interleavings=)` plan-tree fan-out

The architecture is sound and the five-commit layering (kwarg → recording → enumerator → engine → docs) is clean. Two blocking issues below; two non-blocking observations.

---

### B1 — `"critical"` policy produces a duplicate leaf for 2-branch `parallel()` [BLOCKING]

`interleavingCardinality` for `"critical"` returns `2*Branches - 1`. For `Branches=2` that is **3**. The enumerator therefore generates 3 leaves (idx 0, 1, 2). The engine then calls `interleavingOrdering(site, idx)` → `permutationByIndex(2, idx)` for all three.

`permutationByIndex(2, 2)`:
- `fact=[1,1]`, `pool=[0,1]`
- i=0: `digit = 2/1 = 2`; `digit >= len(pool)` → clamp to `1`; `k = 2%1 = 0`; `perm[0] = pool[1] = 1`; `pool=[0]`
- i=1: `digit = 0`; `perm[1] = 0`
- **Result: `[1, 0]`** — identical to idx=1

Leaf 2 silently executes the same branch order as leaf 1. A user who authors `parallel(a, b, interleavings="critical")` sees 3 reported test executions but only 2 distinct orderings. This is the most common case (2-branch parallel is the modal usage).

**Fix options:**
1. Cap `"critical"` cardinality at `min(2*Branches-1, factorial(Branches))` — for 2 branches that's `min(3, 2) = 2`, matching reality.
2. Alternatively, keep cardinality=3 and have `interleavingOrdering` detect and skip duplicates — but that complicates leaf indexing.

Option 1 is the right fix. Same change needed in `interleavingCardinality` and the comment on `interleavingOrdering`.

---

### B2 — Unrecognized kwargs silently accepted in `parallel()` [BLOCKING]

`parseInterleavingsKwarg` reads the `interleavings` key by name via `starKwarg` and returns. Any other kwargs present on the call are silently ignored.

A typo like:
```python
parallel(a, b, interleaving="all")   # missing trailing 's'
```
succeeds without error, producing single-leaf behavior. The author gets no fan-out and no diagnostic.

This is the same silent-failure class that PR #118 fixed for `nondet()` — where unknown kwargs were rejected with `if len(kwargs) > 0 { return nil, fmt.Errorf(...) }`. The fix here is slightly different because `parallel()` now accepts one legitimate kwarg:

```go
// After parseInterleavingsKwarg returns:
for _, kv := range kwargs {
    if k, _ := starlark.AsString(kv[0]); k != "interleavings" {
        return nil, fmt.Errorf("parallel(): unrecognized keyword argument %q", k)
    }
}
```

A test case for `parallel(a, b, interleaving="all")` (typo) expecting a runtime error should accompany the fix.

---

### N1 — `permutationByIndex` comment documents a non-existent fallback [non-blocking]

```
// k must be in [0, n!). Callers that pass an out-of-range index
// get the identity permutation [0..n-1] as a safe fallback...
```

The implementation does NOT return identity for out-of-range `k` — it clamps `digit = len(pool)-1`, producing a clamped (non-identity) permutation. No current caller ever triggers this path (all call sites in `interleavingOrdering` bounds-check first), so there's no correctness bug. But the comment documents a contract that doesn't exist. Correct to: `// out-of-range k is handled by callers; this function's output is undefined for k >= n!.`

---

### N2 — `TestRunAll_LeafDrivenBranchOrder` only guards against panic [non-blocking]

```go
// the fact that both leaves passed (no panic, no nil-deref) under
// the new ordering-driven path is the regression guard.
```

The test doesn't verify that leaf 0 and leaf 1 actually executed in different branch orders. A spec that records branch-call order into a shared `[]int` (mutex-protected) would let the test assert `order[leaf0] != order[leaf1]`. Not required for merge, but worth a follow-up: without this, a regression that always uses identity ordering would still pass the test suite.

---

**Thread creation** in `parallelWithLeaf` (`&starlark.Thread{Name: ...}`) is consistent with `parallelSimple` and `parallelWithExplore` — no issue there.

Fix B1 and B2, then re-submit.

avatar29A:
## Review responses — fixes in c24bb38

Both blockers + both non-blockers fixed in one commit.

### B1 — duplicate critical leaf for 2-branch parallel()

Capped `interleavingCardinality("critical")` and `interleavingOrdering("critical")` at `min(2N-1, N!)`. For N=2 that's `min(3, 2) = 2` — the third leaf no longer exists, so no duplicate. New `TestInterleavings_CriticalLeavesAreDistinct` iterates n=2..4 and asserts every leaf decodes to a unique permutation; broader guard against the same class of bug for any future N.

### B2 — unknown kwargs silently accepted

Added kwarg validator in `builtinParallel` that rejects everything except `"interleavings"`:
```go
for _, kv := range kwargs {
    if k, _ := starlark.AsString(kv[0]); k != "interleavings" {
        return nil, fmt.Errorf("parallel(): unrecognized keyword argument %q", k)
    }
}
```
Mirrors the PR #118 `nondet()` pattern. `TestInterleavings_ParallelRejectsUnknownKwarg` covers the `interleaving="all"` typo case explicitly.

### N1 — wrong permutationByIndex comment

Comment claimed callers got identity for out-of-range k. Implementation actually clamps the Lehmer digit; `interleavingOrdering` bounds-checks before calling in, so no current path is wrong. Corrected the comment to: *"Out-of-range k is the caller's responsibility — clamps to keep the index decode well-defined, but the output is undefined for k ≥ n!."*

### N2 — TestRunAll_LeafDrivenBranchOrder only guarded against panic

Added `TestParallelWithLeaf_LaunchesBranchesInOrder`: calls `parallelWithLeaf` directly with two distinct orderings (`[2,0,1]` and `[1,2,0]`) and verifies the Go-side branch-invocation sequence matches. A regression to identity ordering would now fail this test. Also added `TestParallelWithLeaf_SkipsOutOfRangeIndex` covering the defensive bounds check.

### Test status

- `go test -race ./...` clean
- `go vet ./...` clean
- 4 new test cases + 1 expanded (TestInterleavings_Cardinality gained the 2-branch and 4-branch "critical" rows)
- CI should re-run green

avatar29A (commented):
## Re-review: commit c24bb382 — LGTM

All four items from the initial review are correctly addressed.

**B1 — "critical" duplicate leaf for 2-branch parallel()** ✅  
Both `interleavingCardinality` and `interleavingOrdering` now apply `min(2N-1, N!)`. The factorial is computed once above the switch and caps the critical case in both sites independently. For 2 branches: min(3, 2) = 2. Enumerator and engine use the same formula so they stay in sync. The old dead-code guard (`if c < 2`) is removed correctly — Branches < 2 is already filtered at function entry. `TestInterleavings_CriticalLeavesAreDistinct` iterates n=2..4 and asserts every decoded permutation is unique across all leaves — solid regression cover.

**B2 — Unknown kwargs silently accepted** ✅  
Post-parse loop rejects any kwarg key that is not `"interleavings"`. `TestInterleavings_ParallelRejectsUnknownKwarg` exercises the typo `interleaving="all"` and asserts both the fail outcome and the error message naming the offending kwarg.

**N1 — permutationByIndex comment** ✅  
Corrected: now states the output is undefined for k >= n! and that clamping keeps the decode well-defined (never panics), not that it returns identity. Matches actual behavior.

**N2 — TestRunAll_LeafDrivenBranchOrder weak assertion** ✅  
`TestParallelWithLeaf_LaunchesBranchesInOrder` calls `rt.parallelWithLeaf` directly with two distinct 3-branch orderings and asserts the recorded `seen` slice matches exactly — a regression to always-identity-ordering would fail this. `TestParallelWithLeaf_SkipsOutOfRangeIndex` covers the defensive bounds check with `[-1, 99, 1, 0]`.

**LGTM** — ready to merge.

PR #124, opened 2026-05-26
feat(report): per-leaf axis surfacing in TestResult, manifest, and tests table (A3, RFC-042 §8.10)

avatar29A:
## Summary

Closes A3 — the v0.13.0 Plan-tab integration slice for rc2 fan-out data. Each multi-leaf `TestResult` now carries the leaf's actual axis assignments through to the bundle manifest and the HTML report's tests table.

Before this PR, multi-leaf rows in the tests table read as `test_foo [leaf 0]`, `test_foo [leaf 1]`, ... — you knew leaves existed but couldn't see what each leaf observed without cross-referencing `plan.json`. After: `test_foo [leaf 3 — retries=1, fault=2]` or `test_foo [leaf 5 — wal[10]]` or `test_foo [leaf 1 — spec.star:7#3]`.

## What landed

- **`TestResult`** gains `LeafChoices map[string]int`, `LeafProbabilityOutcomes map[string][]bool`, `LeafInterleavingIDs map[string]int`. Populated by `RunTestLeaf` from the attached PlanLeaf; defensive deep-copy so later mutations don't drift the recorded result.
- **`runTestFanout`** stamps the enumerated leaf 0 axes onto the reused discovery-run result so single-leaf-reuse rows carry the same data as re-executed leaves. Closes the gap where leaf 0 of a choose-only fan-out had nil `LeafChoices`.
- **`bundle.TestRow`** gains `leaf_choices`, `leaf_probability_outcomes`, `leaf_interleavings` (all `omitempty`). `cmd/faultbox`'s `testRowsFromResult` propagates from TestResult.
- **`internal/report/app.js`** gains a `formatLeafAxes(row)` helper returning a compact display string. Tests table's display name now suffixes with axis details when present.

## Scope limit

Per-leaf axes are runtime-observed data and live in **manifest.tests**, not `plan.json`. The Plan tab itself (static analysis, no body run) still renders fault_matrix compositions only. Surfacing rc2 axes in plan.json would require running the body during plan generation; tracked as a follow-up.

## Test plan

- [x] 2 new tests:
  - `TestRunAll_LeafChoicesSurfaceOnTestResult` — 3 leaves of `choose("retries", [0,1,3])` each carry their option index; single-leaf rows omit the field
  - `TestRunAll_LeafInterleavingsSurfaceOnTestResult` — 2 leaves of `parallel(a, b, interleavings="all")` each carry one interleaving entry
- [x] Pre-existing tests unchanged: single-leaf rows still omit the new fields (rc1 manifest shape preserved byte-identically)

## Migration

Additive: existing manifest consumers see the new fields only when they're populated (omitempty). The `leaf_id` shape from rc2 PR #121 is unchanged.

### DIFF
--- a/cmd/faultbox/main.go
+++ b/cmd/faultbox/main.go
@@ -2289,6 +2289,12 @@ func testRowsFromResult(result *star.SuiteResult) []bundle.TestRow {
 			})
 		}
 
+		// Copy the per-leaf axis maps rather than aliasing TestResult's
+		// references. RunTestLeaf already deep-copies them off the
+		// PlanLeaf; we copy here too so a future refactor of
+		// TestResult lifecycle can't introduce a shared-mutation bug
+		// across the manifest row and the suite-result row (review
+		// N2 on PR #124).
 		rows = append(rows, bundle.TestRow{
 			Name:                    tr.Name,
 			Outcome:                 outcome,
@@ -2297,12 +2303,45 @@ func testRowsFromResult(result *star.SuiteResult) []bundle.TestRow {
 			Expectation:             tr.ExpectationName,
 			BypassedRules:           bypassed,
 			LeafID:                  tr.LeafID,
-			LeafChoices:             tr.LeafChoices,
-			LeafProbabilityOutcomes: tr.LeafProbabilityOutcomes,
-			LeafInterleavingIDs:     tr.LeafInterleavingIDs,
+			LeafChoices:             copyIntMapForRow(tr.LeafChoices),
+			LeafProbabilityOutcomes: copyBoolVecMapForRow(tr.LeafProbabilityOutcomes),
+			LeafInterleavingIDs:     copyIntMapForRow(tr.LeafInterleavingIDs),
 		})
 	}
 	return rows
+}
+
+// copyIntMapForRow returns nil for an empty or nil source so the
+// omitempty JSON contract on TestRow.LeafChoices /
+// LeafInterleavingIDs is preserved. Otherwise allocates a fresh map
+// — testRowsFromResult avoids aliasing the SuiteResult's maps so a
+// downstream mutator can't drift both row sets in lockstep (review
+// N2 on PR #124).
+func copyIntMapForRow(m map[string]int) map[string]int {
+	if len(m) == 0 {
+		return nil
+	}
+	out := make(map[string]int, len(m))
+	for k, v := range m {
+		out[k] = v
+	}
+	return out
+}
+
+// copyBoolVecMapForRow mirrors copyIntMapForRow for the per-leaf
+// probability outcomes map. The bool vectors are also deep-copied
+// since they're the load-bearing structure for engine consultation.
+func copyBoolVecMapForRow(m map[string][]bool) map[string][]bool {
+	if len(m) == 0 {
+		return nil
+	}
+	out := make(map[string][]bool, len(m))
+	for k, v := range m {
+		cp := make([]bool, len(v))
+		copy(cp, v)
+		out[k] = cp
+	}
+	return out
 }
 
 // printReplayHints emits a `Replay: faultbox replay …` line per failed
--- a/cmd/faultbox/main.go
+++ b/cmd/faultbox/main.go
@@ -2290,13 +2290,16 @@ func testRowsFromResult(result *star.SuiteResult) []bundle.TestRow {
 		}
 
 		rows = append(rows, bundle.TestRow{
-			Name:          tr.Name,
-			Outcome:       outcome,
-			DurationMs:    tr.DurationMs,
-			Seed:          tr.Seed,
-			Expectation:   tr.ExpectationName,
-			BypassedRules: bypassed,
-			LeafID:        tr.LeafID,
+			Name:                    tr.Name,
+			Outcome:                 outcome,
+			DurationMs:              tr.DurationMs,
+			Seed:                    tr.Seed,
+			Expectation:             tr.ExpectationName,
+			BypassedRules:           bypassed,
+			LeafID:                  tr.LeafID,
+			LeafChoices:             tr.LeafChoices,
+			LeafProbabilityOutcomes: tr.LeafProbabilityOutcomes,
+			LeafInterleavingIDs:     tr.LeafInterleavingIDs,
 		})
 	}
 	return rows
--- a/docs/spec-language.md
+++ b/docs/spec-language.md
@@ -1400,6 +1400,26 @@ fault never fired. The predicate name (e.g. `expect_success`,
 `manifest.tests[].expectation` and surfaces alongside the pill in the
 report's tests table and drill-down header.
 
+#### Per-leaf axis attribution (v0.13.0 rc2)
+
+Multi-leaf rows from `choose()` / `probability=`/ `interleavings=`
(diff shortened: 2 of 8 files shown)


avatar29A (commented):
## Review: A3 — per-leaf axis surfacing (RFC-042 §8.10)

The architecture is clean — three-axis symmetry in `TestResult`, `bundle.TestRow`, and the JS helper; defensive deep-copies in `RunTestLeaf`; the reuse-path stamping for choose-only leaf 0. One blocking gap below.

---

### B1 — No end-to-end test for `LeafProbabilityOutcomes` on `TestResult` [BLOCKING]

The PR adds `LeafChoices`, `LeafProbabilityOutcomes`, and `LeafInterleavingIDs` to `TestResult` and covers all three in the implementation. But the test plan covers only two:

- `TestRunAll_LeafChoicesSurfaceOnTestResult` ✓  
- `TestRunAll_LeafInterleavingsSurfaceOnTestResult` ✓  
- *(probability — absent)*

`TestRunAll_MultiLeafBundleShape` in `probability_fanout_test.go` predates this PR and doesn't assert on `LeafProbabilityOutcomes`. The code path that populates it (`RunTestLeaf` → `copyBoolVecMap(leaf.ProbabilityOutcomes)`) is the most complex of the three (bool-vec deep copy) and is the only one completely untested at the TestResult level.

A test analogous to the choose case — spec with `deny(..., probability=0.5, max_fires=2, mode="exhaustive")`, 4 leaves, assert each `tr.LeafProbabilityOutcomes["<key>"]` is a distinct 2-element vector, and single-leaf rows omit the field — is the missing piece. Should also verify that the two leaves of a combined choose+prob fan-out each carry both maps.

---

### N1 — `leaf_choices` displays option indices, not option values [non-blocking]

`formatLeafAxes` renders `ch[k]` which is the zero-based option *index* into `ChoiceVal.Options`, not the selected *value*. For `choose("retries", [0, 1, 3])`:

- Leaf 0 → `retries=0` (value 0 — coincidentally correct)  
- Leaf 1 → `retries=1` (value 1 — coincidentally correct)  
- Leaf 2 → `retries=2` (display), actual observed value is 3

The design is intentional (option values aren't stored in the manifest — storing the index avoids type-serialisation complexity). But the display is confusing when values and indices diverge. A parenthetical like `retries=2(val=3)` isn't feasible without manifest changes, but the spec-language.md scope-limit note could clarify that the suffix shows option indices. The PR description example `[leaf 3 — retries=1]` happens to be unambiguous only because values and indices align there.

---

### N2 — Shallow map copy in `testRowsFromResult` [non-blocking]

`testRowsFromResult` assigns:
```go
LeafChoices:             tr.LeafChoices,
LeafProbabilityOutcomes: tr.LeafProbabilityOutcomes,
LeafInterleavingIDs:     tr.LeafInterleavingIDs,
```

Both `bundle.TestRow` and the original `TestResult` now share the same map. `RunTestLeaf` uses `copyIntMap`/`copyBoolVecMap` for the same data. In practice this is safe (the results aren't mutated after the rows are built), but the inconsistency is worth noting. If `testRowsFromResult` ever gets a mutation-after-call in a future refactor, this is a latent aliasing bug.

---

### N3 — `TestRunAll_LeafInterleavingsSurfaceOnTestResult` checks count but not distinctness [non-blocking]

```go
if len(tr.LeafInterleavingIDs) != 1 {
    t.Errorf(...)
}
```

Asserts one entry per leaf but not that the two leaves carry *different* index values. A regression where both leaves get index 0 would pass this test. The analogous choose test checks `seen[0]`, `seen[1]`, `seen[2]` are all present. Adding a similar `seen` set for the interleaving indices (expecting {0, 1} across the two leaves) would close the gap.

---

Fix B1 and re-submit.

avatar29A:
## Review responses — fixes in ee4577e

All four items addressed in one commit.

### B1 — missing prob TestResult test

Added two:
- **TestRunAll_LeafProbabilityOutcomesSurfaceOnTestResult** — spec with `deny(probability=0.3, max_fires=2, mode="exhaustive", label="wal")` produces 4 leaves; each carries a distinct 2-bit fire/no-fire vector covering `{00, 01, 10, 11}`. Asserts single-leaf rows omit the field.
- **TestRunAll_CombinedChooseAndProbCarryBothMaps** — composite (choose × probability) test produces 4 leaves; each leaf carries BOTH `LeafChoices` and `LeafProbabilityOutcomes`. Guards against a future refactor that copies one map kind but not the other.

The `copyBoolVecMap` deep-copy path now has end-to-end coverage.

### N1 — option index vs option value clarification

Added a "Per-leaf axis attribution" sub-section in `docs/spec-language.md` after the outcome taxonomy. Explicitly calls out: `retries=2` in the report means "option at index 2"; for `choose("retries", [0,1,3])` that maps to the literal value `3`. The full axis-value spec lives in `plan.json`. Manifest stores indices for type-serialisation simplicity.

### N2 — shallow map copy aliasing

`testRowsFromResult` now deep-copies the three leaf maps via new `copyIntMapForRow` / `copyBoolVecMapForRow` helpers before assigning to `bundle.TestRow`. Returns `nil` on empty input so the `omitempty` JSON contract holds. The double-copy (`RunTestLeaf` → `TestResult`, then `testRowsFromResult` → `TestRow`) is intentional: the row boundary is where independence becomes load-bearing.

### N3 — interleaving distinctness assertion

`TestRunAll_LeafInterleavingsSurfaceOnTestResult` now asserts the two leaves carry distinct interleaving indices ({0, 1}, not duplicates), and that they agree on the site key. A regression to always-index-0 now fails this test.

### Test status

- `go test -race ./...` clean
- 2 new tests + 1 strengthened
- CI should re-run green

avatar29A (commented):
## Re-review: commit ee4577e9 — LGTM

All four items from the initial review are correctly addressed.

**B1 — LeafProbabilityOutcomes uncovered** ✅  
Two new tests added: `TestRunAll_LeafProbabilityOutcomesSurfaceOnTestResult` (4-leaf probability fan-out, asserts all four bit-vectors {00, 01, 10, 11} appear across leaves + single-leaf omit) and `TestRunAll_CombinedChooseAndProbCarryBothMaps` (choose × probability cross-product, each leaf asserted to carry both maps). The `copyBoolVecMap` deep-copy path now has end-to-end coverage at the TestResult level.

**N1 — leaf_choices shows indices not values** ✅  
New "Per-leaf axis attribution (v0.13.0 rc2)" section in `spec-language.md` explicitly calls out the index semantics with the concrete example: `choose("retries", [0,1,3])` leaf 2 displays `retries=2` (index), observed value is 3. Full axis-value spec lives in `plan.json`.

**N2 — Shallow copy in testRowsFromResult** ✅  
`copyIntMapForRow` and `copyBoolVecMapForRow` helpers added in `cmd/faultbox/main.go`. Return `nil` for empty/nil input so the `omitempty` JSON contract is preserved — consistent with the `if len(...) > 0` guards in `RunTestLeaf`.

**N3 — TestRunAll_LeafInterleavingsSurfaceOnTestResult weak** ✅  
Now tracks `seenIdx` across both leaves and asserts the set covers `{0, 1}`. Consistent parallel-site key across leaves is also verified. A regression to always-index-0 would fail.

**LGTM** — ready to merge.

PR #125, opened 2026-05-26
feat(rfc-043): per-leaf assume + AST denylist + Q2/Q3 polish (rc2)

avatar29A:
## Summary

Completes RFC-043 rc2 — wires the four operators into the plan-tree fan-out engine that landed in A1/A2 and addresses Q2/Q3 from PR #118's review.

## What lands

**Per-leaf `assume=` evaluation (RFC-043 §5.4 rc2).** Per-test predicates now see the current leaf's axis assignment at body entry, including body-time `choose()` calls.
- `currentChoicesDict` consults `rt.currentLeaf` via `ChoiceVal.Selected`
- `rt.planAxes` persists the choose-axis schema across leaves so the dict can be rebuilt after `rt.choices` is reset for the next leaf's body run
- `runTestFanout` re-executes leaf 0 for choose-only fan-out (was only for prob/par)
- `PlanLeaf.IsDiscovery` flag marks the synthetic leaf0 used for discovery; assume= eval is skipped on that pass so predicates indexing body-time keys don't spuriously error before any axis has been recorded

**§8.7 AST denylist for assume() (`assume_sandbox.go`, new ~150 LOC).** Mirrors `monitorSandboxDenylist`. Rejects predicates referencing `fault*`/`service*`/`parallel`/`halt`/`eventually`/`always`/`monitor`/`await_*`/`assert_*`. Hooked into `LoadString`/`LoadFile` alongside the existing monitor sandbox. Limitation: named `def` predicates slip past — same as the monitor predicate limitation.

**Q3 — predicate Starlark errors → `Result="error"` (was `"fail"`).** A KeyError or AttributeError inside a predicate is a spec-authoring bug, not a SUT behavioral failure. CI gates that distinguish the two now get accurate signal.

**Q2 — normalize `Mode=""` → `"exhaustive"` at parse time when `max_fires > 0`.** Internal representation matches the documented default; downstream code can use `Mode == "exhaustive"` equality rather than the looser `Mode != "stochastic"`.

## Test plan

- [x] 4 new tests + 1 updated:
  - `TestAssume_PerLeafChoicesVisible` — 4-leaf cross-product, predicate halts 2 leaves (retries=7), 2 pass
  - `TestAssume_SandboxRejectsFaultCall` — lambda referencing `fault()` rejected at spec load
  - `TestAssume_SandboxRejectsTopLevelAssumeWithDeniedBuiltin` — top-level `assume()` also goes through the sandbox
  - `TestAssume_SandboxAllowsChoicesDictReads` — `choices.get()` and dict reads not blocked
  - `TestAssume_PredicateInternalError` updated: now expects `Result="error"` per Q3
- [x] Pre-existing assume tests continue to pass

## Docs

- RFC-043 status flipped from rc1 → rc1+rc2
- `docs/nondeterministic-operators.md` — visibility/sandbox callouts rewritten
- CHANGELOG paragraph extended
- Feature manifest row updated

## Deferred (tracked)

- **§8.5 plan-walker-time `assume=` pruning** — today predicates halt at body entry (Result="halted") but the plan tree still contains the pruned leaf. Lifting pruning into the enumerator to shrink cardinality up front is a follow-up.
- **§8.6 cost-guard for fan-out budgets.**


avatar29A (commented):
Reviewed commit `a069962`. Four changes in scope: per-leaf `assume=` evaluation (§5.4 rc2), §8.7 AST denylist (`assume_sandbox.go`), Q3 error classification, Q2 mode normalization.

Core implementation is correct. No blocking bugs. Four non-blocking issues below.

---

## Non-blocking

### N1 — Dead `else` branch in `runTestFanout` (cleanup, confusing)

After the PR's change the second condition's `else` is structurally unreachable. The guard:

```go
if len(axes) == 0 && len(probAxes) == 0 && !hasActiveParAxes {
    return []TestResult{tr0}
}
```

…exits early when all three axes are empty. The later condition:

```go
if len(axes) > 0 || len(probAxes) > 0 || hasActiveParAxes {
    // re-run leaf 0
} else {
    // Reuse the discovery run … [tr0 stamped with axes]
}
```

…is the logical complement of the guard, so the `else` is never reached. The reuse path was live before this PR (choose-only fan-out fell into it); after extending the `if` condition to also include `len(axes) > 0`, the guard now always fires first for that case too. Dead code that describes the old choose-only reuse contract will mislead the next reader. Remove the `else` block.

### N2 — Duplicate comment block before the `if len(axes) > 0 || …` condition

Two back-to-back comment blocks describe the same if-condition:

```go
// Leaf 0's discovery run used the RNG / single-ordering path (no
// PlanLeaf attached). Re-run leaf 0 with the explicit
// zero-vector / zero-ordering assignment so attribution is
// consistent across all leaves whenever any fan-out kind is
// active.
// Re-execute leaf 0 with explicit assignment when ANY fan-out
// axis is present. Previously this only triggered for prob/par
// fan-out …
if len(axes) > 0 || len(probAxes) > 0 || hasActiveParAxes {
```

The first block is the pre-rc2 comment, the second is the rc2 replacement. They overlap. Merge into one.

### N3 — `parseProbabilityFanoutKwargs` doc comment is stale after Q2

The function-level doc still says:

```
//   - mode must be "exhaustive" or "stochastic"; empty string maps to
//     "" (caller treats as default = exhaustive in v0.13.0).
```

After Q2 `mode=""` + `maxFires > 0` normalizes to `"exhaustive"` at parse time. The "caller treats" note is now wrong — the caller sees `"exhaustive"` directly. Update the doc to state the post-normalization invariant: `Mode == "exhaustive"` whenever `MaxFires > 0` and no explicit `mode="stochastic"` was provided.

### N4 — No explicit test for Q2 normalization

No test asserts `FaultDef.Mode == "exhaustive"` when only `max_fires= N` is passed to `delay()` / `deny()`. The existing probability fan-out tests exercise the normalized code path, but none verify the internal representation. Add a unit test in `builtins_determinism_test.go` (or adjacent) along the lines of:

```go
func TestParseProbabilityFanout_DefaultModeNormalized(t *testing.T) {
    rt := New(testLogger())
    err := rt.LoadString("spec.star", `
service("svc", addr="localhost:9000")
fault("svc", "read", probability=0.5, max_fires=2)
`)
    if err != nil {
        t.Fatalf("LoadString: %v", err)
    }
    // Verify the FaultDef's Mode was normalized to "exhaustive"
    // (RFC-043 Q2). Without this the downstream equality-check
    // Mode == "exhaustive" would silently miss default-mode faults.
    fd := rt.LookupFaultDef("svc", "read")
    if fd.Mode != "exhaustive" {
        t.Errorf("Mode = %q, want exhaustive (Q2 normalization)", fd.Mode)
    }
}
```

(Adjust API to whatever the test helper surface exposes for `FaultDef` lookup.)

---

The core mechanics — `planAxes` schema persistence, `IsDiscovery` skip, `c.Selected(leaf)` per-leaf dispatch, `validateAssumeLambdasInSource` denylist, Q3 `"error"` classification — are all correct and well-tested. `TestAssume_PerLeafChoicesVisible` closes the critical rc2 path.

avatar29A:
## Review responses — fixes in d193d81

All four non-blockers addressed.

### N1 — dead else branch removed

The choose-only reuse path was unreachable: the no-axis short-circuit above returns before we get to the if/else, and by the time we enumerate leaves at least one axis is in play. Removed the dead else block and simplified the if to an unconditional re-execute of leaf 0. `tr0` is now a discovery-only result whose recordings drive enumeration; its value is consumed by the planner above and the canonical leaf-0 result comes from the re-execute.

### N2 — duplicate comments merged

The two stacked comment blocks (old "RNG / single-ordering path" + new "RFC-043 §5.4 rc2") collapsed into one explanation that names the IsDiscovery skip and per-leaf assume= as the reasons.

### N3 — `parseProbabilityFanoutKwargs` doc updated

Comment now describes Q2 normalization correctly:
- omitted `mode=` with `max_fires>0` → normalizes to `"exhaustive"`
- omitted `mode=` with no `max_fires` → stays `""` (no-fan-out path)
- `mode="exhaustive"` without `max_fires` → rejected (B4 from PR #121)

### N4 — explicit Mode normalization test

`TestProbabilityFanout_ModeParses` gains two table-driven rows:
- `deny("EIO", probability=0.5, max_fires=2)` → `Mode="exhaustive"`
- `deny("EIO", probability=0.3, max_fires=5)` → `Mode="exhaustive"`

Pins the internal representation directly so a future refactor breaking the normalization fails the test instead of silently changing semantics downstream.

### Test status

- `go test -race ./internal/star/...` clean
- Full suite green
- CI should re-run green on the push

avatar29A (commented):
Verified fix commit `d193d81`. N1–N4 all addressed correctly.

- **N1**: dead `else` removed; unconditional leaf-0 re-execute with `_ = tr0` making the discard intent clear
- **N2**: two stacked comment blocks merged into one coherent explanation
- **N3**: `parseProbabilityFanoutKwargs` doc now correctly describes Q2 normalization (omitted `mode=` + `max_fires > 0` → `"exhaustive"`; no-`max_fires` stays `""`) and the B4 rejection case
- **N4**: `TestProbabilityFanout_ModeParses` gains two explicit Q2 rows asserting `Mode == "exhaustive"` when only `max_fires=` is provided

LGTM.

PR #127, opened 2026-05-28
refactor(plan): unified NonDeterministicChoice interface (RFC-044 §8.2 / C2)

avatar29A:
## Summary

C2 — unified fan-out machinery (RFC-044 §8.2). The three plan-tree axis kinds (`ChoiceVal`, `ProbFaultSite`, `ParallelSite`) now implement a single `NonDeterministicChoice` interface; `enumerateLeaves` is a thin wrapper over a generic mixed-radix walker.

Pure refactor — no user-facing API change, all pre-existing testops goldens byte-identical.

## What lands

- **`NonDeterministicChoice` interface** (`internal/star/nondet.go`, new):
  ```go
  type NonDeterministicChoice interface {
      Cardinality() int
      Apply(leaf *PlanLeaf, digit int)
  }
  ```
- **Three impls.** `*ChoiceVal`, `ProbFaultSite`, `ParallelSite` each get a 5-10 line `Cardinality()` + `Apply()` pair. Compile-time witnesses pinned in tests.
- **`collectChoices(axes, probAxes, parAxes)`** flattens the three source slices into a homogeneous `[]NonDeterministicChoice` in source order, filtering zero-cardinality axes (anonymous choose(), single-option axes, no-max_fires prob sites, single-policy parallel sites).
- **`expandLeaves(choices)`** is the generic mixed-radix walker. One pass over the slice, calls `Cardinality()` once per axis to size the total, then `Apply(digit)` per (leaf, axis) pair.
- **`enumerateLeaves`** keeps its (axes, probAxes, parAxes) signature — now a one-liner: `return expandLeaves(collectChoices(...))`. The dozen existing callers don't churn.

## Scope honesty

The original RFC said "five separate code paths" but the post-rc2 reality was only one (`enumerateLeaves`) — choose / probability / interleaving all already went through it. `fault_matrix` is intentionally NOT a `NonDeterministicChoice`: it produces discrete *test* entries at spec load time via `internal/plan/`, not *leaves* of one test at body time. The RFC §8.2 status header now documents this scope correction.

## Test plan

- [x] Testops goldens (`TestGoldens/*`) unchanged — plan-tree output byte-identical
- [x] 6 new unit tests:
  - Per-interface contract for each of the three impls
  - `collectChoices` preserves source order + filters zero-cardinality
  - End-to-end cross-product (2 × 3 × 4 × 6 = 144 leaves) with index/assignment invariants
  - Empty input produces one degenerate leaf

## Out of scope

- §8.6 `observe.*` and §8.7 `decoder()` unifications — tracked as C3, independent slice
- Tutorial chapters for RFC-042/043 — C4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,20 +84,26 @@ default. **Deferred:** §8.5 plan-walker-time `assume=` pruning
 follow-up) and §8.6 cost-guard. User guide:
 [docs/nondeterministic-operators.md](docs/nondeterministic-operators.md).
 
-**RFC-044 — spec language simplification (partial).** First slice
-(§8.1 + §8.3 + §8.4 + §8.5) — the docs-and-deprecations bundle:
-RFC-013 (`param()`) formally **withdrawn** and the
-`docs/rfcs/0013-parameterized-scenarios.md` header updated with a
-rationale paragraph pointing users to RFC-043 `choose("name",
-[opts])`; RFC-002 (`domain()`) formally **withdrawn** via a new
-`docs/rfcs/0002-domain.md` stub; `faultbox generate` deprecated in
-favor of `faultbox plan --suggest` with a one-time stderr warning
-and a CLI-reference callout (removal in v0.14.0); a new
-"Feature interactions — what caps below L4" subsection in
+**RFC-044 — spec language simplification (§8.1 + §8.2 + §8.3 +
+§8.4 + §8.5).** C1 shipped the docs-and-deprecations bundle: RFC-013
+(`param()`) formally **withdrawn** (superseded by RFC-043
+`choose("name", [opts])`); RFC-002 (`domain()`) formally
+**withdrawn** via a new `docs/rfcs/0002-domain.md` stub; `faultbox
+generate` deprecated in favor of `faultbox plan --suggest` with a
+one-time stderr warning (removal in v0.14.0); a "Feature
+interactions — what caps below L4" subsection in
 `docs/spec-language.md` documents the determinism ceilings for
-`service(remote=…)`, `service(reuse=True, …)`, and `mock_service(…)`.
-Remaining RFC-044 sections (§8.2 unified fan-out machinery, §8.6
-`observe.*`, §8.7 `decoder()`) tracked as follow-up slices.
+`service(remote=…)` and `service(reuse=True, …)`. **C2** ships the
+unified fan-out machinery (§8.2): the three plan-tree axis kinds
+(`ChoiceVal`, `ProbFaultSite`, `ParallelSite`) now implement a
+single `NonDeterministicChoice` interface in
+`internal/star/nondet.go` with `Cardinality()` / `Apply(leaf, digit)`
+methods; `enumerateLeaves` is a thin wrapper over a generic
+mixed-radix walker `expandLeaves`. Adding a new axis kind in the
+future requires only implementing the two-method contract — no
+per-kind branching in the enumerator. All pre-existing testops
+goldens unchanged. Remaining RFC-044 sections (§8.6 `observe.*`,
+§8.7 `decoder()`) tracked as follow-up slices.
 RFC-046 carries the post-L1 roadmap (gVisor Path B/C, L4 hermetic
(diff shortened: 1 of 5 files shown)


avatar29A (commented):
Reviewed commit `84a3d11`. Pure refactor — verified the mixed-radix equivalence fully.

**Correctness verdict: byte-identical to pre-refactor.**

Equivalence trace:
- `collectChoices` output order — `[*ChoiceVal..., ProbFaultSite..., ParallelSite...]` — matches the three sequential loops in the old `enumerateLeaves` exactly.
- Each `Cardinality()` returns the same `n` the old per-kind code computed: `len(Options)` / `1<<MaxFires` / `interleavingCardinality(p)`.
- `expandLeaves` does `rem % card` / `rem /= card` in the same left-to-right order, producing the same mixed-radix digit per (leaf, axis).
- `ProbabilityOutcomes` and `InterleavingIDs` nil-vs-present: old code set them only when the respective map was non-empty (`if len(probs) > 0`); new `Apply` lazy-allocates only when called; `collectChoices` gates inclusion on `Cardinality() > 0` — same result.
- Degenerate leaf: both old and new return `{Index:0, Choices:map[string]int{}}` for the no-axes case.

No blocking bugs. Two non-blocking issues below.

---

## Non-blocking

### N1 — `expandLeaves` has a silent wrong-output case for zero-cardinality inputs

The degenerate-leaf guard is:

```go
if len(choices) == 0 {
    return []PlanLeaf{{Index: 0, Choices: map[string]int{}}}
}
```

If `choices` contains a zero-cardinality element the guard doesn't fire, `total` folds to 0, and `make([]PlanLeaf, 0)` returns an empty slice — no degenerate leaf. `collectChoices` prevents this in the normal path, but the function has no precondition comment or assertion to communicate the invariant. The `Cardinality()` docstring says "Must return ≥ 1 for the axis to participate" — echoing that in `expandLeaves` would close the gap:

```go
// precondition: all elements have Cardinality() > 0;
// collectChoices guarantees this for the standard call path.
```

### N2 — `TestNonDet_ExpandLeavesIdenticalToEnumerateLeaves` omits `ProbabilityOutcomes` spot-check

The test verifies `Choices` and `InterleavingIDs` for the boundary leaves but not `ProbabilityOutcomes`. For leaf 143: digit-for-prob = 23 % 4 = 3 → binary 11 → `["wal"] = [true, true]`. One assertion completes the "byte-identical" safety-net claim:

```go
wantProb := []bool{true, true}
if !reflect.DeepEqual(got[wantLen-1].ProbabilityOutcomes["wal"], wantProb) {
    t.Errorf("leaf %d prob = %v, want %v", wantLen-1,
        got[wantLen-1].ProbabilityOutcomes["wal"], wantProb)
}
```

---

Interface design, scope correction in the RFC, compile-time witnesses, and the six tests are all solid.

avatar29A:
## Review responses — fixes in 43f0509

Both non-blockers addressed.

### N1 — expandLeaves precondition

Added an explicit **Precondition** paragraph to the doc comment: every element of `choices` must have `Cardinality() ≥ 1`. The canonical entry point `collectChoices` already enforces this by filtering zero-cardinality axes; the comment pins the contract for any future caller that builds the slice by hand.

### N2 — ProbabilityOutcomes assertion

`TestNonDet_ExpandLeavesIdenticalToEnumerateLeaves` now asserts on the all-zero and all-max leaves' `ProbabilityOutcomes` vectors too:
- Leaf 0 wal vector → `[false, false]` (digit 0)
- Leaf 143 wal vector → `[true, true]` (digit 3, binary 11)

The byte-identical claim now covers all three `PlanLeaf` fields end-to-end.

avatar29A (commented):
Verified fix commit `43f0509`. N1 and N2 both correct.

- **N1**: `expandLeaves` precondition documented with the zero-cardinality edge case explained and the `collectChoices` safe path called out.
- **N2**: Leaf-0 (`[false, false]`) and leaf-143 (`[true, true]`) `ProbabilityOutcomes` assertions land alongside the existing `Choices`/`InterleavingIDs` checks — all three PlanLeaf fields now verified for both boundary leaves.

LGTM.

PR #128, opened 2026-05-28
refactor(spec-lang): observe.* + decoder("name") unifications (RFC-044 §8.6/§8.7 / C3)

avatar29A:
## Summary

C3 — namespace unifications, the final RFC-044 slice. v0.13.0 introduces the `observe` Starlark struct (§8.6) and the unified `decoder("name", ...)` dispatcher (§8.7). Pre-rc2 top-level builtins remain as deprecated aliases that emit a one-time stderr warning per process and route through the canonical implementation. Removal of the aliases is scheduled for v0.14.0.

## What lands

**§8.6 — `observe.*`** (`internal/star/observe_decoder.go::makeObserveModule`)
- `observe.stdout` / `observe.stderr` attributes on a `starlarkstruct.Struct`.
- Legacy `stdout()` / `stderr()` route through `builtinStdoutDeprecated` / `builtinStderrDeprecated` — one-time warning + delegate to the same value-producing source as the namespace.
- **Scope correction:** the RFC originally listed `topic()`, `tail()`, `wal_stream()`, `poll()` as additional factories. Those were never top-level builtins (they ship only as event-source plugins consumed via `observe=` lists on `service()`); only `stdout` / `stderr` needed unifying. RFC §8.6 status note records the correction.

**§8.7 — `decoder("name", ...)`** (`builtinDecoder`)
- Dispatches `"json"` / `"logfmt"` / `"regex"` (with `pattern=` kwarg). All four routes converge in `builtinDecoder` so `DecoderVal` construction has a single source of truth.
- Legacy `json_decoder()` / `logfmt_decoder()` / `regex_decoder()` keep working with one-time warnings; old implementations removed (no callers).

**Deprecation warning infrastructure**
- `warnDeprecated(oldName, replacement)` emits one stderr line per (process, name). Suppression is per-name so ten `stdout()` calls produce one warning. Per-process rather than per-spec — CI runs typically load one spec; the user mental model is "the migration is the same regardless of which spec triggered it."
- `resetDeprecationWarnings()` exposed for tests.

## Test plan

- [x] 7 new tests:
  - `TestObserve_ModuleExposesStdoutAndStderr` — `observe.stdout()` produces `*ObserveSourceVal`
  - `TestObserve_StdoutDeprecatedAliasStillWorks` — legacy form keeps type identity
  - `TestDecoder_UnifiedDispatcher` — three names + regex's `pattern=` kwarg
  - `TestDecoder_Rejections` — bad arity / wrong type / unknown name / missing pattern / disallowed kwargs (7 cases)
  - `TestDecoder_LegacyAliasesStillWork` — three legacy builtins still produce `DecoderVal`
  - `TestObserve_DeprecationWarning` — captures stderr; asserts one warning per legacy builtin even when called multiple times; warning names the replacement
  - `TestDecoder_DeprecationWarning` — same for the three legacy decoder builtins

## Docs

- `spec-language.md` decoder table gains a "Legacy alias (deprecated, removed in v0.14.0)" column; example code switches to `observe.stdout(decoder=decoder("json"))` form; migration callout below.
- RFC-044 status header flipped to fully Implemented; §8.6 and §8.7 status notes with the scope correction.
- CHANGELOG paragraph extended.
- Feature-manifest rows updated.

## v0.13.0 epic close-out

This completes RFC-044. With C3 in, all five v0.13.0 RFCs (040, 041, 042, 043, 044) have landed end-to-end:
- **RFC-040** Determinism levels — L1 contract with `unmediated_io` events
- **RFC-041** Temporal properties — `eventually` / `always` / `await_*` / `monitor` / `test()`
- **RFC-042** Exploration plan + body re-execution engine + interleaving execution
- **RFC-043** Non-deterministic operators — `choose` / `nondet` / `halt` / `assume` with per-leaf fan-out
- **RFC-044** Spec language simplification — withdrawals + unified fan-out machinery + namespace unifications

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,8 +102,17 @@ methods; `enumerateLeaves` is a thin wrapper over a generic
 mixed-radix walker `expandLeaves`. Adding a new axis kind in the
 future requires only implementing the two-method contract — no
 per-kind branching in the enumerator. All pre-existing testops
-goldens unchanged. Remaining RFC-044 sections (§8.6 `observe.*`,
-§8.7 `decoder()`) tracked as follow-up slices.
+goldens unchanged. **C3** ships the namespace unifications
+(§8.6 + §8.7): a new `observe` Starlark struct exposes
+`observe.stdout` and `observe.stderr` as attributes; a unified
+`decoder("name", ...)` dispatcher handles `"json"`, `"logfmt"`,
+and `"regex"`. The pre-rc2 top-level `stdout()` / `stderr()` /
+`json_decoder()` / `logfmt_decoder()` / `regex_decoder()` builtins
+remain registered as deprecated aliases that emit a one-time
+stderr warning per process and route through the canonical
+implementation — DecoderVal construction has a single source of
+truth, so a future decoder bug fix lands in one place. Removal
+of the aliases is scheduled for v0.14.0.
 RFC-046 carries the post-L1 roadmap (gVisor Path B/C, L4 hermetic
 mode, L5 instruction-boundary research).
 
--- a/docs/feature-manifest.md
+++ b/docs/feature-manifest.md
@@ -128,8 +128,8 @@ Protocol-level fault proxy rewrites wire-level responses. Critical because this 
 
 | Feature | Tier | Mechanism | Status | Notes |
 |---|---|---|---|---|
-| Event sources (stdout, wal_stream, topic, tail, poll) | 2 | `internal/eventsource/*_test.go` | 🟢 | |
-| Decoders (json, logfmt, regex) | 2 | `internal/eventsource/decoder` unit tests | 🟢 | |
+| Event sources (stdout, wal_stream, topic, tail, poll) | 2 | `internal/eventsource/*_test.go` + `internal/star/observe_decoder_test.go` (RFC-044 `observe.*` namespace + deprecated aliases) | 🟢 | RFC-044 §8.6: top-level `stdout()` / `stderr()` deprecated → `observe.stdout` / `observe.stderr`; one-time stderr warning per process, removal v0.14.0. |
+| Decoders (json, logfmt, regex) | 2 | `internal/eventsource/decoder` unit tests + `internal/star/observe_decoder_test.go` (unified dispatcher + deprecated aliases) | 🟢 | RFC-044 §8.7: unified `decoder("name", ...)` dispatcher; `json_decoder()` / `logfmt_decoder()` / `regex_decoder()` deprecated, removal v0.14.0. |
 | Container reuse (`reuse=True`) | 2 | No coverage | 🔴 | |
 | Lima VM dev environment | 3 | Manual (`make env-verify`) | 🟡 | |
 
--- a/docs/rfcs/0044-spec-language-simplification.md
+++ b/docs/rfcs/0044-spec-language-simplification.md
@@ -1,6 +1,6 @@
 # RFC-044: Spec Language Simplification
 
-> **Status: Implemented partial (v0.13.0).** §8.1 (RFC-013 withdrawal) + §8.3 (`faultbox generate` deprecation) + §8.4 (RFC-002 withdrawal) + §8.5 (L1-cap documentation) shipped in C1 (#126). §8.2 (unified fan-out machinery) shipped in C2 — the three plan-tree fan-out axis kinds (`ChoiceVal`, `ProbFaultSite`, `ParallelSite`) now implement a single `NonDeterministicChoice` interface (`internal/star/nondet.go`); `enumerateLeaves` is a thin wrapper over a generic mixed-radix walker. **Deferred:** §8.6 `observe.*` and §8.7 `decoder()` module unifications.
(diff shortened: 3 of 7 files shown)


avatar29A (commented):
**Principal Engineer Review — RFC-044 C3 / PR #128**

One blocking doc/code inconsistency, three non-blocking observations.

---

**B1 — `spec-language.md` says \"per spec load\", implementation is \"per process\" [BLOCKING]**

The migration callout in `docs/spec-language.md` reads:

> *…emit a one-time stderr warning **per spec load**…*

But the implementation in `observe_decoder.go` is explicit:

```go
// Per-process (not per-spec) is intentional: most CI runs load
// exactly one spec, so per-process matches the user mental model.
// A test harness that LoadStrings multiple specs in sequence sees
// one warning total
```

`deprecationWarnings` is a package-level `map[string]bool{}` that persists for the process lifetime. A test harness that runs three specs in sequence sees the warning once, not three times. That's intentional — and it's documented correctly in the code comment and PR description — but the user-facing doc contradicts both. Fix: change \"per spec load\" to \"per process\" in the callout.

---

**N1 — Tombstone comment in `builtins.go` [non-blocking]**

The four-line archaeology block at the old location of the three decoder builtins:

```go
// (Pre-RFC-044 decoder builtins lived here. They were folded into
// the unified builtinDecoder dispatcher in observe_decoder.go; …)
```

The git log carries this story. The comment adds noise without explaining a non-obvious invariant. Drop it.

---

**N2 — `makeObserveModule()` comment claims constructor name is \"observe\"; it isn\'t [non-blocking]**

```go
// The struct constructor name is "observe" so error messages
// reading `<observe.stdout>` are clear about the namespace.
func makeObserveModule() *starlarkstruct.Struct {
    return starlarkstruct.FromStringDict(
        starlarkstruct.Default,   // ← "struct", not "observe"
```

`starlarkstruct.Default` is the constant `starlark.String("struct")`. Error messages from the struct itself will read `struct has no .foo field`, not `observe has no .foo field`. The builtin names (`"observe.stdout"`, `"observe.stderr"`) do appear in call-site errors, which is the actually-valuable part, but the comment\'s constructor-name claim is factually wrong. Either switch to `starlarkstruct.Constructor("observe")` to match the claim, or rewrite the comment to just explain the builtin naming.

---

**N3 — `json_decoder(foo="bar")` silently changed from accepted to error; untested [non-blocking]**

Pre-C3, `builtinJSONDecoder` and `builtinLogfmtDecoder` accepted and ignored arbitrary kwargs (no kwarg-checking code). Now the deprecated aliases delegate to `builtinDecoder` which has `if len(kwargs) > 0 { return nil, fmt.Errorf(...) }` for `"json"` and `"logfmt"`. So `json_decoder(foo="bar")` now returns an error where it previously succeeded silently. This is the right final behavior, and no real users exist to break, but `TestDecoder_LegacyAliasesStillWork` only exercises the happy path — a negative case like `json_decoder(foo="bar")` → error would make the behavioral change explicit and intentional rather than accidental.

---

Core design — `makeObserveModule`, `builtinDecoder`, `warnDeprecated`/`resetDeprecationWarnings`, `captureStderr` test helper — all clean. `expandLeaves` delegation from C2 is untouched and correct. Seven tests cover the surface well.

Fix B1 and the PR is good to merge.

avatar29A:
## Review responses — fixes in 759e36a

B1 blocker + all three non-blockers addressed.

### B1 — "per spec load" → "per process"

Doc text in `spec-language.md` migration callout corrected to match the implementation: warnings fire once **per process**, not per spec load. Added the "test harness loading multiple specs sees the warning once total" clarification.

### N1 — tombstone comment removed

The "Pre-RFC-044 decoder builtins lived here" block in `builtins.go` is now gone. `git log` and the new code in `observe_decoder.go` are the sources of truth.

### N2 — `observe` constructor identity

Switched `makeObserveModule` from `starlarkstruct.Default` (which is literally `starlark.String("struct")`) to `starlark.String("observe")` directly. The value now prints as `observe(...)` rather than `struct(...)` — namespace identity is visible at every surface that prints a Starlark value.

### N3 — explicit negative case for legacy alias strictness

New `TestDecoder_LegacyAliasesRejectUnknownKwargs` pins the behavior change:
- `json_decoder(foo="bar")` → fails (was: silently accepted pre-rc2)
- `logfmt_decoder(pattern="x")` → fails (was: silently accepted)
- `regex_decoder()` → fails with "requires pattern"

Comment on the test explains this is intentional: legacy aliases route through `builtinDecoder`, which enforces the new contract. A future compatibility patch loosening this requires an explicit decision.

### Test status

- `go test -race ./...` clean
- 1 new test added (8 total for C3 now)
- CI should re-run green

avatar29A (commented):
LGTM. All four findings addressed in 759e36a5 — B1 doc fix, N1 tombstone removed, N2 constructor switched to `starlark.String("observe")` with corrected comment, N3 kwarg-rejection behavior pinned by `TestDecoder_LegacyAliasesRejectUnknownKwargs` with contract note. RFC-044 fully shipped.

PR #126, opened 2026-05-27
chore(rfc-044): C1 — RFC-013/RFC-002 withdrawals + generate deprecation + L1-cap docs

avatar29A:
## Summary

LGTM on PR #125 — B (RFC-043 rc2 wiring) merged as `083f264`. This PR opens **C1**, the first slice of RFC-044 (spec language simplification).

C1 is docs-and-deprecations: no engine changes. Locks the low-risk surface before the heavier refactor slices (C2 unified fan-out machinery, C3 observe.* / decoder() unification) land.

## What lands

- **§8.1 — RFC-013 (`param()`) withdrawn.** `param()` and `choose()` were the same primitive with two names; v0.13.0-rc2 made `choose("name", [...])` the unified surface. RFC-013 was never shipped, so no code change. Status header updated with rationale.
- **§8.4 — RFC-002 (`domain()`) withdrawn.** Early multi-service-grouping sketch; project pivoted to `service()`-first and never came back. New `docs/rfcs/0002-domain.md` stub explains the pivot. RFC-004 and RFC-005 reference RFC-002; they're themselves draft/deferred and will re-anchor when rewritten.
- **§8.3 — `faultbox generate` deprecated.** Command continues to work but prints a one-time stderr warning pointing users to `faultbox plan --suggest`. Removal scheduled for v0.14.0. CLI reference updated.
- **§8.5 — L1-cap feature documentation.** New "Feature interactions — what caps below L4" subsection in `docs/spec-language.md` tabulates the determinism ceilings for `service(remote=...)`, `service(reuse=True, ...)`, and `mock_service(...)`. Code-side error wiring is unreachable today (L2..L5 already error at spec load); the docs are the user-facing contract until L4 itself lands.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./...` clean
- [x] No new tests — the slice is documentation-only except for one stderr warning in `generateCmd`

## Out of scope for C1

- **§8.2 — Unified fan-out machinery.** ~1-2 weeks refactor consolidating five plan-tree expansion paths into one `NonDeterministicChoice` type. Tracked as **C2** — independent epic.
- **§8.6 — `observe.*` unification.** Wraps existing event-source builtins with deprecated aliases. Tracked as **C3**.
- **§8.7 — `decoder("name", ...)` unification.** Same shape as §8.6 for json/logfmt/regex decoders. Tracked as **C3**.
- **Tutorial chapters for RFC-042/043.** Tracked as **C4** (the deferred original A1 work).

Issue #122, opened 2026-05-26
bug: probability fan-out site order is non-deterministic across runs


avatar29A:
## Issue

The recording loop in `applyFaults` iterates `map[string]*FaultDef`, which is unordered in Go. When a spec declares ≥ 2 probability faults on the same service, `recordProbFault` appends sites in random order across runs.

`enumerateLeaves` uses the recorded site order to assign mixed-radix digits to leaves — so the same spec produces stable cardinality but **unstable LeafID-to-axis mapping** between runs. Worse: `faultbox plan` and `faultbox test` consulting the same spec at different times can disagree on which leaf is which.

Pre-existing (the original recording code had the same map range), confirmed by principal-engineer reviewer on PR #121 follow-up.

## Repro

```python
svc = service("svc", image="busybox", cmd=["sh","-c","sleep 1"])

# Two probabilistic axes on the same service — non-deterministic order.
fault(svc, write=deny("EIO", probability=0.3, max_fires=2, label="wal"))
fault(svc, read=deny("EIO", probability=0.3, max_fires=2, label="cache"))
```

Leaf 1 might pin `wal=fired, cache=not_fired` on one run and the reverse on another.

## Fix

`sort.Slice` the recorded sites by `Key` before passing to `enumerateLeaves`, OR record into a sorted slice from the start. Single-source-of-truth ordering for the plan walker.

## Related

- PR #121 follow-up review
- RFC-042 §8.9
- Affects `faultbox plan` vs `faultbox test` cross-validation in CI gates

PR #129, opened 2026-05-28
docs(tutorial): RFC-042/043 chapters in Part 4 (C4)

avatar29A:
## Summary

C4 — the deferred A1 tutorial work. Two new chapters in Part 4 (Safety & Verification) cover the v0.13.0-rc2 fan-out vocabulary end-to-end. No code changes.

## What lands

**[Chapter 25 — Non-deterministic Operators (`choose` / `assume` / `halt` / `nondet`)](docs/tutorial/04-safety/25-choose-and-assume.md)**
- "One test, 18 leaves" motivating problem (3-axis matrix)
- `choose("name", [opts])` named vs anonymous; option-index-vs-value display caveat
- `assume()` per-test + top-level, sandbox semantics, predicate Starlark errors → `Result="error"` (Q3 polish)
- `halt(reason)` body-time pruning + rejection contexts + halt-vs-assume comparison
- `nondet()` shorthand + the `nondet(svc)` arity-disambiguation
- The discovery-run mental model — why body runs N+1 times for N leaves
- Worked composition example, 4 exercises

**[Chapter 26 — Plan-Tree Fan-Out](docs/tutorial/04-safety/26-plan-fanout.md)**
- Probability fan-out: 2^N leaf table, per-occurrence vector, full mode-rules matrix, migration story
- `parallel(interleavings=)` policy taxonomy, reserved values, launch-ordering scope limit
- `faultbox plan` CLI: text/JSON output, `--check-cost --max-instances N` gate
- Composition: 72-leaf cross-product example
- Per-leaf trace navigation in the HTML report
- "When to use which axis kind" decision table, 4 exercises

Tutorial README updated with both chapters slotted into Part 4 alongside Ch 24 (Determinism).

## Test plan

- [x] `go test ./...` clean — `docs/` package's markdown link checker covers the new files
- [x] All cross-references resolve (Ch 4, Ch 14, Ch 25→26, Ch 5)
- [x] Code examples typecheck mentally against the rc2 surface (named choose, assume= kwarg, max_fires, interleavings=)

## v0.13.0 epic close-out

With C4 in, the v0.13.0 epic is done:
- 5 RFCs end-to-end shipped (040, 041, 042, 043, 044)
- 9 PRs merged this session (#121 #123 #124 #125 #126 #127 #128 plus this one)
- 2 follow-up issues filed (#119 RunTest select race, #122 non-deterministic prob site order)
- Two new tutorial chapters teaching the rc2 operator + fan-out vocabulary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,20 @@ stderr warning per process and route through the canonical
 implementation — DecoderVal construction has a single source of
 truth, so a future decoder bug fix lands in one place. Removal
 of the aliases is scheduled for v0.14.0.
+
+**Tutorial — RFC-042/043 chapters.** Part 4 (Safety & Verification)
+gains two new chapters covering the v0.13.0-rc2 fan-out vocabulary:
+**Ch 25 — Non-deterministic Operators (`choose` / `assume` /
+`halt` / `nondet`)** walks the four RFC-043 operators end-to-end
+from the "one test, 18 leaves" motivating problem through the
+sandbox semantics and the discovery-run mental model; **Ch 26 —
+Plan-Tree Fan-Out** covers the RFC-042 plan tree, probability
+fan-out with `max_fires=`/`mode=`, `parallel(interleavings=)`, the
+`faultbox plan --check-cost` gate, and the cross-product
+composition rules. Both chapters include exercises and the
+`Result="error"` vs `"fail"` distinction for predicate Starlark
+errors.
+
 RFC-046 carries the post-L1 roadmap (gVisor Path B/C, L4 hermetic
 mode, L5 instruction-boundary research).
 
--- a/docs/tutorial/04-safety/25-choose-and-assume.md
+++ b/docs/tutorial/04-safety/25-choose-and-assume.md
@@ -102,7 +102,7 @@ test("checkout",
 )
 ```
 
-The `assume=` kwarg on `test()` takes a list of predicates. Each predicate receives a `choices` dict mapping axis names to the leaf's selected option index. Leaves where any predicate returns False are recorded as **halted** — they don't contribute to pass/fail.
+The `assume=` kwarg on `test()` takes a list of predicates. Each predicate receives a `choices` dict mapping axis names to the leaf's selected option **value** (not the index — the dict carries what the body would observe from `choose()`, so a predicate that reads `choices["retries"]` against `choose("retries", [0, 1, 3])` sees one of `0`, `1`, `3` directly). Leaves where any predicate returns False are recorded as **halted** — they don't contribute to pass/fail.
 
 > **Visibility:** Per-test `assume=` predicates evaluate at body entry. They see the leaf's axis assignment for any `choose()` call recorded during the discovery run (the first body execution; see below). Conditional choose() calls — axes that only appear in some branches — may not show up in `choices` for every leaf. Best practice: declare all axes unconditionally at the start of the body.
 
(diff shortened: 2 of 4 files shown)


avatar29A (commented):
**Principal Engineer Review — RFC-042/043 tutorial / PR #129**

One blocking factual error, one non-blocking arithmetic mistake.

---

**B1 — Ch 25: `choices` dict contains option values, not option indices [BLOCKING]**

Under the `assume(predicate)` section:

> Each predicate receives a `choices` dict mapping axis names to the leaf's selected **option index**.

And in the Visibility note:

> They see the leaf's axis assignment for any `choose()` call recorded during the discovery run … Conditional choose() calls … may not show up in `choices` for every leaf.

Both uses of "index" are wrong. `currentChoicesDict()` in `internal/star/assume.go` sets dict values via `c.Selected(leaf)`, which returns `c.Options[idx]` — the actual Starlark value from the options list. For `choose("retries", [0, 1, 3])` on the third leaf, `choices["retries"]` is the integer `3`, not the option index `2`.

The predicate examples in the chapter are already correct (they do arithmetic on values), so only the descriptive labels need fixing:

- "mapping axis names to the leaf's selected **option index**" → "mapping axis names to the leaf's selected **option value**"
- The Visibility note's phrasing is fine; no change needed there.

Note: the "Leaf-ID convention" callout *under `choose()`* correctly says the HTML **report display** shows the option index — that reads from `PlanLeaf.Choices[name]` (the digit). The report and the predicate dict are two different things: the report shows the index, the predicate receives the value. The callout is right; only the assume() section description is wrong.

---

**N1 — Ch 26 Exercise 3: "four-times-larger" should be "eight-leaf" [non-blocking]**

> Add `max_fires=3, mode="exhaustive"` and observe the **four-times-larger** plan tree.

`max_fires=N` → `2^N` leaves. A stochastic baseline (no `max_fires`) produces 1 leaf per run; `max_fires=3` produces `2³ = 8` leaves — 8× larger, not 4×. (The chapter's own probability table shows the formula; a reader who checks the math will find the discrepancy.) Fix: "observe the eight-leaf plan tree" (or drop the multiplier claim entirely and just say "observe how much larger the plan tree becomes").

---

Everything else is solid — the discovery-run mental model section in Ch 25 is particularly well explained, the halt-vs-assume comparison table is exactly the right framing, and the mode-rules matrix in Ch 26 covers all the Q2/B4 edge cases correctly. Fix B1 and this is good to merge.

avatar29A:
## Review responses — fixes in 250f3d8

Both findings addressed.

### B1 — Ch 25 `choices` dict labels

Corrected the descriptive prose: the dict carries the option **value** (what the body observes from `choose()`), not the index. Added an explicit example: `choices["retries"]` against `choose("retries", [0, 1, 3])` returns one of `0` / `1` / `3` directly.

The predicate code examples were already correct — they read the dict expecting integer values, not indices. The bug was the descriptive label only.

Worth noting the distinction from `bundle.TestRow.leaf_choices` documented in `spec-language.md`, which IS the option index (it's the underlying `PlanLeaf.Choices` map). That doc stays correct as-is; the two surfaces serialize different things.

### N1 — Ch 26 Exercise 3 multiplier

Replaced "the four-times-larger plan tree" with "the eight-leaf plan tree (2³)" — concrete and verifiable from `faultbox plan` output. 2³ = 8 leaves from a 1-leaf stochastic baseline.

`go test ./docs/...` clean.

avatar29A (commented):
LGTM. Both findings addressed in 250f3d8d — B1 choices dict label corrected to option value with the concrete [0,1,3] example inline, N1 multiplier replaced with "eight-leaf plan tree (2³)". v0.13.0 epic closed.

PR #130, opened 2026-05-29
fix: deterministic prob site order (#122) + regression test for #119 select-race drain

### LINKED ISSUE #122: bug: probability fan-out site order is non-deterministic across runs

avatar29A:
## Summary

Single PR addressing both follow-up issues filed during the v0.13.0 epic.

- **Closes #122** — probability fan-out site order is now deterministic across runs. `applyFaults` iterates the faults map in sorted-syscall order so `faultbox plan` and `faultbox test` produce identical LeafID → axis mappings for the same spec.
- **Reinforces #119** — the bodyDone-arm select drain fix from PR #120 (already in main) gains the regression test it never had. A future refactor that removes the drain now fails deterministically.

## Note on #119

While drafting this PR I confirmed #119 was actually closed earlier this session via PR #120 — the bodyDone arm at `runtime.go:1511` already contains the non-blocking drain on `alwaysFired` / `twFired`. What was missing was an end-to-end test pinning the behavior. This PR adds that test alongside the #122 fix, so both follow-ups are addressed in the single PR you requested.

### #122 fix (`internal/star/runtime.go`)

`applyFaults` previously iterated `map[string]*FaultDef` to record probability fan-out sites. Map range is unordered in Go, so two probabilistic axes on the same service produced sites in random order across runs. Mixed-radix LeafID assignment is stable in cardinality but unstable in axis attribution.

```go
syscalls := make([]string, 0, len(faults))
for s := range faults {
    syscalls = append(syscalls, s)
}
sort.Strings(syscalls)
for _, syscall := range syscalls {
    fd := faults[syscall]
    // ... existing recording logic ...
}
```

Inline comment cites the issue and explains the cross-tool agreement requirement.

### #119 regression test

`TestIssue119_BodyDoneDrainsAlwaysFired` — registers a spec with `body=lambda: None`, `expect=always(lambda t: False)`, `timeout=1s`. The body returns immediately, alwaysFired closes from the first event, and the drain MUST observe the violation and convert `cause` to `TerminationImmediateFail`. Without the drain (regression), the bodyDone arm passes through with `cause=Natural` and the test would report `Result="pass"` instead of `"fail"`.

## Test plan

- [x] 2 new tests:
  - `TestIssue122_ProbFaultSiteOrderIsDeterministic` — 50 fresh-runtime iterations on a 3-axis spec; asserts byte-identical site order every run
  - `TestIssue119_BodyDoneDrainsAlwaysFired` — pins the drain behavior; regression-only

## Impact

- **#122** is a CI gate fix: specs using `--check-cost` could see false-positive cost changes across runs because mixed-radix expansion attributed differently. Sorted iteration makes the plan walker fully deterministic.
- **#119** test is regression-only — no behavior change.

### DIFF
--- a/internal/star/determinism_fixes_test.go
+++ b/internal/star/determinism_fixes_test.go
@@ -82,12 +82,6 @@ log_d = deny("EIO", probability=0.3, max_fires=2, mode="exhaustive", label="log"
 // regression that removed the drain would fail deterministically.
 func TestIssue119_BodyDoneDrainsAlwaysFired(t *testing.T) {
 	rt := New(testLogger())
-	// always(lambda t: False) registered as a per-test expect makes
-	// the first event arrival close alwaysFired. The body does
-	// nothing visible — but await_event below blocks long enough
-	// for the synthetic event we emit afterwards to land before
-	// body return, so by the time bodyDone is ready, alwaysFired
-	// is also ready.
 	src := `
 def body():
     pass
--- a/internal/star/determinism_fixes_test.go
+++ b/internal/star/determinism_fixes_test.go
@@ -1,0 +1,123 @@
+package star
+
+import (
+	"context"
+	"strings"
+	"testing"
+)
+
+// TestIssue122_ProbFaultSiteOrderIsDeterministic — regression for
+// issue #122 (probability fan-out site order is non-deterministic
+// across runs).
+//
+// Before the fix, applyFaults iterated faults (a map[string]*FaultDef)
+// in Go's randomized order, so two probabilistic axes on the same
+// service produced sites in random order across runs. The plan
+// walker uses site order to assign mixed-radix digits to leaves —
+// unstable order → stable cardinality but unstable LeafID → axis
+// mapping. The fix sorts the syscall keys before iteration.
+//
+// This test runs the recording path many times in a single process
+// and asserts the site sequence is byte-identical every time.
+func TestIssue122_ProbFaultSiteOrderIsDeterministic(t *testing.T) {
+	src := `
+svc = service("svc", image="busybox", cmd=["sh","-c","sleep 1"])
+wal_d = deny("EIO", probability=0.3, max_fires=2, mode="exhaustive", label="wal")
+cache_d = deny("EIO", probability=0.3, max_fires=2, mode="exhaustive", label="cache")
+log_d = deny("EIO", probability=0.3, max_fires=2, mode="exhaustive", label="log")
+`
+	var firstKeys []string
+	for i := 0; i < 50; i++ {
+		rt := New(testLogger())
+		if err := rt.LoadString("spec.star", src); err != nil {
+			t.Fatalf("LoadString: %v", err)
+		}
+		// Drive the recording path: simulate the body's fault() call
+		// installing three rules on one service. applyFaults takes the
+		// nil-session bail-out path because we don't start the service.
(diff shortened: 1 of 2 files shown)


avatar29A (commented):
**Principal Engineer Review — #122 fix + #119 regression test / PR #130**

No blocking issues. One stale comment to drop.

---

**#122 fix** (`runtime.go`) — correct and minimal. `sort.Strings` on the extracted syscall keys before the recording loop is exactly the right fix; the inline comment names the issue and explains why cross-tool LeafID agreement requires stable ordering. `sort` already imported; diff is complete.

**#119 regression test** — the test design is sound. `always(lambda t: False)` fires on the first event arrival, closing `alwaysFired`; the bodyDone drain then converts `cause` to `TerminationImmediateFail`; `Result="fail"` is the correct observable. A future removal of the drain regresses to `"pass"` and the assertion catches it deterministically.

---

**N1 — stale comment block in `TestIssue119_BodyDoneDrainsAlwaysFired` [non-blocking]**

Lines 87–90 of the test (inside the function body, just before the spec string):

```go
// The body does nothing visible — but await_event below blocks long enough
// for the synthetic event we emit afterwards to land before
// body return, so by the time bodyDone is ready, alwaysFired is
// also ready.
```

The spec is just `def body(): pass`. There is no `await_event`, no synthetic event, and `RunConfig{}` has no special setup. The comment describes an earlier draft of the test. The real mechanism (always(False) fires on the first real event; the drain is what matters) is already explained in the doc-comment above the function. Drop the four inner lines.

---

Fix N1 and ship.

avatar29A:
## Review response — fix in 01b3a0d

N1 addressed. Removed the four-line stale comment from inside `TestIssue119_BodyDoneDrainsAlwaysFired` — leftover from an earlier draft where I'd considered an `await_event`-based race forcing before settling on the simpler `always(False)` + no-op body design. The outer doc-comment already explains the real mechanism.

avatar29A (commented):
LGTM. N1 fixed in 01b3a0dc — stale inner comment block dropped, outer doc-comment is the sole explanation. Both issues closed.

